PR #1, opened 2020-10-02
Adding the basic repo architecture

pechouc:
The aim of this PR is to add a basic architecture to the repository. It is a mix of what was already in the repo (LICENSE, README.md and .gitignore) and of a Python package template, slightly adapted following [this Slack thread](https://data-for-good.slack.com/archives/C01BB6H16FM/p1601497845010600).


VincGargasson (dismissed):
Looks like to be a good base to start to work on. :)

requirements.txt
```diff
@@ -0,0 +1,10 @@
+pip>=9
+setuptools>=26
+wheel>=0.29
```
frgfm, line 3: Are those really required? They are always expected to be up to date but comes as base python setup

requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+pytest
+pandas
+numpy
+dash
+dash-core-components 
+dash-html-components
+plotly
```
frgfm, line 10: Better version these requirements (at least a lower bound) ;)

Makefile
```diff
@@ -0,0 +1,60 @@
...
+	@rm -fr pyronear-webapp-*.dist-info
+	@rm -fr pyronear-webapp.egg-info
+
+install:
+	@pip install . -U
+
+all: clean install test black check_code
+
+
+uninstal:
```
frgfm, line 34: small typo over there

frgfm (changes requested):
Thanks for the PR! Looks good to me, apart from `.DS_Store` files and the requirements (lower bound versioning at least). Would you mind removing them please? (better add them into the `.gitignore` to avoid this later on)

requirements.txt
pechouc: `pip` and `setuptools` do but I am not sure for `wheel`.

Typically, the logs of CI checks show that `wheel` was downloaded and installed based on the requirements (requirement was not already satisfied). 

I am removing the two first ones, what do you think about `wheel`? Would you still remove it?

Makefile
pechouc: Nice catch, thanks! It is corrected.

requirements.txt
frgfm: No need to bother, we can check later on, I was mostly curious :sweat_smile:

requirements.txt
pechouc: I added the versions I usually use as a lower bound

pechouc:
I have pushed a new commit that reflects your comment and re-requested your review but please let me know if that's not the process to follow 😅

frgfm (changes requested):
Thanks for the quick edits! One last thing: you missed the `.github/.DS_Store` file ;)

frgfm (approved):
Thanks a lot for the PR and the edits!

frgfm:
Oops I updated the repo settings! For next PRs, it's better to use "Squash & merge" rather than basic Merge. The reason behind this is that it enforces each commit on master to have passed the tests :+1:

pechouc:
Very clear! Sorry about that, I'll keep that in mind for next time 😊

PR #7, opened 2020-10-14
graphs.py new file + some changes in homepage.py layout !

nrslt:
Hi guys,
I just added a graphs.py file that will hold all plotly figure objects generated for various indicators (depending on the DS team: meteorological data, risk evaluation, historical information ...)
I also modified the general layout of the homepage: since we expect to find a large enough map when landing on homepage + possibility to slide/change map view + graphs, that represented a lot of information so I decided to put the graphs at the bottom of the page for ease of reading

let me know what you think, of course the layout can be modified and I let @pechouc add the slider


nrslt:
I just did the small modifications for the flake8 checks to go smoothly!

app/homepage.py
```diff
@@ -34,51 +34,49 @@
...
-        dbc.Col(dcc.Dropdown(id='user_department_input',
-                            options=[
-                            {'label': 'Ardèche', 'value': 'Ardèche'},
-                            {'label': 'Gard', 'value': 'Gard'},
-                            {'label': 'Landes', 'value': 'Landes'}
-                            ]))
-        ]),
-    dbc.Row(
+         dbc.Col(dcc.Dropdown(id='user_department_input',
+                              options=[{'label': 'Ardèche', 'value': 'Ardèche'},
```
VincGargasson, line 40: @nrslt  You will include later the logic that zoom on the department selected by the user?

VincGargasson (changes requested):
LGTM ! @nrslt could you add a pic to your PR to see what it looks like while running? thanks !

app/homepage.py
nrslt: @VincGargasson definitely! As I explained yesterday, the container/row/column logic that is inherent of [dash bootstrap components](https://dash-bootstrap-components.opensource.faculty.ai/) is really practical and easy to remodel.
Here is what it looks like at the moment:
Capture d’écran, le 2020-10-15 à 08 30 45
Capture d’écran, le 2020-10-15 à 08 31 13

nrslt: @VincGargasson As for the department zoom, for now I just added the same map as alerts.py. We will need to add a callback element that links the user choice of department with a new map, centered on that department.

Again, the idea here is more about quickly setting a template for us to keep on working on the homepage (for @pechouc to work on the slider, which has its own available slot, and for you to work on the map)
Let me know if this logic makes sense and if you'd like any modification before approving this PR!

Capture d’écran, le 2020-10-15 à 08 38 46


cheers

nrslt (commented):
so if it's ok with you, I propose we set this basic layout in order to all have a common ground to work on!

pechouc (approved):
I might join the conversation a bit late but I agree with the idea of setting a preliminary layout and build upon it in parallel streams 😊

nrslt (commented):
@VincGargasson is it ok with you?

Akilditu (approved):
@nrslt thanks for the PR ! It's indeed an interesting baseline to experience pages structure.

VincGargasson (approved):

PR #10, opened 2020-10-31
feat: adding a new 'one-pager' dashboard in homepage and revamping navbar

Akilditu:
### Aim of the Pull Request

Hi Everyone

This PR aims at : 

- Gathering both alerts and risks dashboards into a main one on homepage and initiate new interaction abilities
- Revamping Navbar design

### More detailed info 

The Navbar now incorporates PyroNear logo with a black background, a new CTA style and a centered text retreiving the username logged : 

Capture d’écran 2020-10-31 à 19 58 26

Both risks and alerts maps are gathered on the same dashboard, with a filter control box on the left with the ability to : 

- Switch from Alerts to Risks maps
- Apply the Satellite or "Schématique" layers to both maps
- Filter opacity for risks map (filter already existing on /risks page)

demo_pyro

As you may notice meteo graphs under the map are not displayed anymore but this can be done setting the `meteo_graphs()` `display` param to `True` in `homepage.py`

Thanks for reviewing 👍


frgfm:
@Akilditu small consideration: there will be some assets that we have to put in the repository, but for logos of the organizations, I would suggest that there are hosted remotely and that our different apps only use the URL.

app/homepage.py
```diff
@@ -14,73 +14,131 @@
...
+                         id='map_style_button')
+
+    return html.Center(button)
+
+
+# This function takes as input the number of clicks on the button defined above and returns the layer style to use
+def choose_map_style(n_clicks):
+
+    # Because we start with the alerts map, if the number of clicks is even, this means that
+    # we are still using the alerts map and we may want to risk one
```
pechouc, line 44:
```suggestion
    # we are still using the alerts map and we may want to switch to the risks one
```

app/homepage.py
```diff
@@ -14,73 +14,131 @@
...
+    else:
+        return ''
+
+
+# ------------------------------------------------------------------------------
+# Content and App layout
 
-# instantiating the map object from alerts.py: to be modified depending on
-# what we expect the user to view first on homepage
+# Instantiating the alert map object from alerts.py and setting it a the default map object:
```
pechouc, line 81:
```suggestion
# Instantiating the alert map object from alerts.py and setting it as the default map object:
```

app/homepage.py
```diff
@@ -14,73 +14,131 @@
...
 
-body = dbc.Container(
-    [dbc.Row(
-        [dbc.Col(html.H1('Bienvenue sur Pyronear')),
-         dbc.Col(dcc.Dropdown(id='user_department_input',
-                              options=[{'label': 'Ardèche', 'value': 'Ardèche'},
-                                       {'label': 'Gard', 'value': 'Gard'},
-                                       {'label': 'Landes', 'value': 'Landes'}
-                                       ]))
+# Instantiating map layers button object from alters.py
```
pechouc, line 90:
```suggestion
# Instantiating map layers button object from alerts.py
```

app/homepage.py
```diff
@@ -14,73 +14,131 @@
...
+map_layers_button = build_layer_style_button()
+
+# Instantiating map style button object
+map_style_button = build_map_style_button()
+
+# Instantiating meteo graphs, set to True to display them under the map, Fale to hide them
```
pechouc, line 96:
```suggestion
# Instantiating meteo graphs, set to True to display them under the map, False to hide them
```

app/navbar.py
```diff
@@ -7,24 +7,58 @@
...
-                                 dbc.DropdownMenuItem('Niveaux de Risque', href='risks')],
-                             )],
-        brand='À propos',
-        brand_href='/home',
-        sticky='top')
+            dbc.DropdownMenuItem('Alertes et Infrastructures', href='alerts'),
+            dbc.DropdownMenuItem('Niveaux de Risque', href='risks'),
+        ],
+        label="Tableaux de bord",
+        className="ml-auto flex-nowrap mt-3 mt-md-0 btn-group dropleft",
```
pechouc, line 32: I am not sure to understand what the `className` defined here (and I see a similar one in the instantiation of the `user_item` object) is designed for and where it comes from. Could you please describe briefly its use or point us to any related resource? Many thanks!

app/homepage.py
```diff
@@ -14,73 +14,131 @@
...
-body = dbc.Container(
-    [dbc.Row(
-        [dbc.Col(html.H1('Bienvenue sur Pyronear')),
-         dbc.Col(dcc.Dropdown(id='user_department_input',
-                              options=[{'label': 'Ardèche', 'value': 'Ardèche'},
-                                       {'label': 'Gard', 'value': 'Gard'},
-                                       {'label': 'Landes', 'value': 'Landes'}
-                                       ]))
+# Instantiating map layers button object from alters.py
+map_layers_button = build_layer_style_button()
```
pechouc, line 91: This comment is more of a question I am asking myself than a change request. Would it be cleaner to directly call the functions instead of their output when building the `body` object? 

This might not be clear said this way but so far, we generally create a new variable (here `map_layers_button`) which simply contains the output of the related function (`build_layer_style_button`) and then, we call it without any particular transformation in the app layout block (at line 123 in this case). Equivalently, we could directly call the function at line 123 and skip the creation of the object. 

I don't know what the best practice is here and I don't have any strong opinion, but I guess that we would spare quite a few lines and variables by doing so, which can only make our code more readable, right?

pechouc (approved):
Thanks a lot @Akilditu for this PR!

I have left a few comments, mostly typos and two questions, but everything seems up and running, so I approved it 😊

Akilditu:
> @Akilditu small consideration: there will be some assets that we have to put in the repository, but for logos of the organizations, I would suggest that there are hosted remotely and that our different apps only use the URL.

I agree @frgfm , we could consider adding this logo version to the pyronear.github.io repo so that it will be hosted on pyronear.org. What do you think ?

app/navbar.py
Akilditu: Hi @pechouc here some explanations about those bootstrap css classes : 

- ml-auto : margin left auto in regards of other columns on the same row
- flex-nowrap force inline alignment as explained [here](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap)
- mt et mt-md for margin top 
- btn-group dropleft to ensure that the dropdown will occur one the left (to match responsive needs) as detailed [here](https://getbootstrap.com/docs/4.0/components/dropdowns/)

Hope that helps

app/homepage.py
Akilditu: Totally agreed, to be honest i followed this 'instantiation logic' to ensure code unity with respect to what has been done so far :)

I think that doing a proper PR for this refactoring task would be a good way to deal with it right ?

app/homepage.py
Akilditu: Thanks, I will fix all this missprints in the next commit

app/homepage.py
pechouc: Yes, I guess this logic is inherited from the code initially provided with the tutorials. And fully agreed with the idea of a separate PR!

frgfm:
> I agree @frgfm , we could consider adding this logo version to the pyronear.github.io repo so that it will be hosted on pyronear.org. What do you think ?

Yes, sounds like our best option here!

frgfm:
Here you go @Akilditu : https://pyronear.org/img/logo_letters_orange.png

Akilditu:
> Here you go @Akilditu : https://pyronear.org/img/logo_letters_orange.png

Yes, will commit this and then it will be ready to merge 👍

Akilditu:
@frgfm, @pechouc I think that we're good to merge ?

frgfm:
> @frgfm, @pechouc I think that we're good to merge ?

I'll let @pechouc give his input on this, as I only took a look at the logo part!
One small thing to consider later: we might have to consider multilanguage options (i.e. having text content stored somewhere in different languages) at some point. Perhaps we should check what our options are here to avoid having heavy design constraints later on (in another issue/PR of course).

Akilditu:
> One small thing to consider later: we might have to consider multilanguage options (i.e. having text content stored somewhere in different languages) at some point. Perhaps we should check what our options are here to avoid having heavy design constraints later on (in another issue/PR of course).

@frgfm This could indeed be and interesting feature to have multilang but I m not sure about the priority of this. I actually think that the next big step in term of design / functional constraints will be determined by the interviews to come with firemen. 
As they are the main users of this platform and mostly french I guess we could consider this more as a long-term priority.

VincGargasson (approved):
Same as PE, as it works and we already discussed it, I have nothing special to add. 
Will see in another PR how we determine if `display` is true or not (button for user?).
Anyway, thanks for huge improvement Seb 👍

PR #13, opened 2020-11-07
Code Refactoring

### LINKED ISSUE #12: Light Code Refactoring
# Starting point of the issue

> This comment is more of a question I am asking myself than a change request. Would it be cleaner to directly call the functions instead of their output when building the `body` object? 
> 
> This might not be clear said this way but so far, we generally create a new variable (here `map_layers_button`) which simply contains the output of the related function (`build_layer_style_button`) and then, we call it without any particular transformation in the app layout block (at line 123 in this case). Equivalently, we could directly call the function at line 123 and skip the creation of the object. 
>
> I don't know what the best practice is here and I don't have any strong opinion, but I guess that we would spare quite a few lines and variables by doing so, which can only make our code more readable, right?

_Originally posted by @pechouc in https://github.com/pyronear/pyronear-webapp/pull/10#discussion_r515630455_

# Scope of the issue 

Following the quick discussion mentioned above, the aim of this issue is to weigh the pros and the cons of a light refactoring of the code that we have written so far. The main goal of the edits would be to eliminate the instantiations we are currently making and could avoid in files like ```alerts.py```, ```risks.py``` and ```homepage.py```.

# Why do such edits? 

The fact that we systematically instantiate each object in the code creates multiple variables that could be avoided by directly calling the functions instead. 

Avoiding these instantiations would allow us to:

* Clarify the code since for each component of the web-application, we have at the moment: a function name, a variable name and an ```id``` used in the callbacks. Sometimes, they are the same and in other cases, they differ which can bring confusions in both situations.

* Keep standalone functions. To give an example of this issue, one can look at the ```AlertsApp()``` for instance: the function uses several variables that have been defined outside of it (```nav```, ```space```, ```layer_style_button```...). 

* Spare a few lines of code. 

# Other remarks

* It seems that this practice of systematically instantiating the objects derives from the [code](https://github.com/joelsewhere/dash_app) that was provided with the tutorial that we followed initially to build the structure of the app. 

* Example of a chunk of code that would be edited:

```
# ------------------------------------------------------------------------------
# App layout (finally!)

# Instantiating the navigation bar
nav = Navbar()

# Adding a first separator between the navigation bar and the button
space = dcc.Markdown('---')

# Instantiating the button to change the map layer style
layer_style_button = build_layer_style_button()

# Adding a second separator between the button and the map
separator = dcc.Markdown('---')

# Finally instantiating the map object
map_object = build_alerts_map()


# Gathering all these elements in a HTML Div and having it returned by the AlertsApp function
def AlertsApp():
    layout = [nav,
              space,
              layer_style_button,
              separator,
              map_object]

    return html.Div(layout)
```

* It could give, after the refactoring:

```
# ------------------------------------------------------------------------------
# App layout (finally!)

def AlertsApp():
    layout = [Navbar(),   # Instantiating the navigation bar
              dcc.Markdown('---'),   # Adding a first separator between the navigation bar and the button
              build_layer_style_button(),   # Instantiating the button to change the map layer style
              dcc.Markdown('---'),   # Adding a second separator between the button and the map
              build_alerts_map()]   # Finally instantiating the map object

    return html.Div(layout)
```

# Question

Let me know what you think of this proposed modification! Do we all agree on this change? Do you fear it might make the code less explicit?

pechouc:
# Aim of the PR

This Pull Request follows up on and closes #12: it aims at refactoring the code we have written so far, mainly to avoid instantiating objects outside of the functions that then call them. Some changes in the comments are also proposed, especially in the ```alerts.py``` file, where code blocks have been slightly reorganised. 

# Changes Made

* The main goal of the edits is to eliminate the instantiations we are currently making and could avoid in files like ```alerts.py```, ```risks.py``` and ```homepage.py```.

* To better describe the goal of the modifications, here is the example that was provided in the issue:

```
# ------------------------------------------------------------------------------
# App layout (finally!)

# Instantiating the navigation bar
nav = Navbar()

# Adding a first separator between the navigation bar and the button
space = dcc.Markdown('---')

# Instantiating the button to change the map layer style
layer_style_button = build_layer_style_button()

# Adding a second separator between the button and the map
separator = dcc.Markdown('---')

# Finally instantiating the map object
map_object = build_alerts_map()


# Gathering all these elements in a HTML Div and having it returned by the AlertsApp function
def AlertsApp():
    layout = [nav,
              space,
              layer_style_button,
              separator,
              map_object]

    return html.Div(layout)
```

* It gives, after the refactoring:

```
# ------------------------------------------------------------------------------
# App layout (finally!)

def AlertsApp():
    layout = [Navbar(),                             # Instantiating the navigation bar
              dcc.Markdown('---'),               # Adding a first separator between the navigation bar and the button
              build_layer_style_button(),       # Instantiating the button to change the map layer style
              dcc.Markdown('---'),               # Adding a second separator between the button and the map
              build_alerts_map()]                  # Finally instantiating the map object

    return html.Div(layout)
```

# Second Commit 

As noticed by @abdi-adel, there was no way to come back to the home page from "specialised" dashboards via the navigation bar. The goal of the second commit is to correct this impossibility by adding the home page to the dropdown menu of the navigation bar (and to correct a typo).

### DIFF
--- a/app/homepage.py
+++ b/app/homepage.py
@@ -62,7 +62,8 @@ def meteo_graphs(display=False):
     if display is True:
         return dbc.Row(
             [
-                dbc.Col([html.H2("Données météorologiques"), dcc.Graph(figure=meteo_fig)],
+                # In the following line, we instantiate the Plotly Graph Objects figure storing the graphs
+                dbc.Col([html.H2("Données météorologiques"), dcc.Graph(figure=generate_meteo_fig())],
                         md=4),
                 dbc.Col([html.H2("another indicator"),
                         dcc.Graph(figure={"data": [{"x": [1, 2, 3], "y": [1, 4, 9]}]})],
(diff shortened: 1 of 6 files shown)


app/homepage.py
```diff
@@ -77,70 +78,54 @@ def meteo_graphs(display=False):
...
+                            {'label': 'Gard', 'value': 'Gard'},
+                            {'label': 'Landes', 'value': 'Landes'}],
+                        placeholder="Départements"),
+                    md=3),
+            ]
+        ),
+        dbc.Row(
+            [dbc.Col(
+                [
+                    dcc.Markdown('---'),
```
VincGargasson, line 106: This part of the code has been changed too in my PR. I guess the first PR that will be merged won't have to deal with conflicts 😅

app/utils.py
```diff
@@ -27,21 +27,30 @@ def get_info(feature=None, feature_type=None):
...
-    # if no object are hovered, it just return standard statement
+
+    # If no object are hovered, it just return standard statement
     return header_dept + [html.P('Faites glisser votre curseur sur un département')]
 
 
 def build_info_object(app_page):
-
+    """
+    This function takes as input the page of interest and outputs the relevant
```
VincGargasson, line 48: As the number of function we have in this script is growing fast, I'm wondering if we should to put set standard in our function's docstrings.
Like this for example: 
1. Brief description of the function
2. Argument (Name + type)
3. Return output

Maybe this would require a specific PR to update it in the whole app.

VincGargasson (approved):
Thanks PE for this PR.
LGTM!  Aligned with what you mentioned in your issue. 
This is clearer.

pechouc:
Thanks a lot for your review, comment and approval @VincGargasson!

> This part of the code has been changed too in my PR. I guess the first PR that will be merged won't have to deal with conflicts 😅

I guess I was going to merge this one now but do let me know if you prefer that we first move forward with yours! I haven't had the time to review it yet but I could definitely make it a priority. And in any case, if we merge mine beforehand, don't hesitate to ping me could I be of any help to manage the conflicts that will appear 😊

> As the number of function we have in this script is growing fast, I'm wondering if we should to put set standard in our function's docstrings. Maybe this would require a specific PR to update it in the whole app.

I couldn't agree more! And I am also aligned with the need for a specific PR to update it in the whole app. 

Once this is done, I think that there are tools that can help us generate a clean documentation out of the docstrings in the code, which could be useful as well while scripts become more and more complex.

pechouc:
Merged following the discussion above and after oral exchange with @VincGargasson and @Akilditu 😊

PR #15, opened 2020-11-09
Adding markers' popup and video display function

VincGargasson:
## **GOAL**

For the user :
1. Showing markers popup with coordinate of the marker (for now)
2. Integrate a function to show video when clicking on the marker (a Simpson one for now :))

For us (dev): 

1. Creation of `user_selection_area()` function to be able to rebuild the left sidebar dash component several time in the app.

### Code structure
 We have quite a few function and It can be sometime complicated to know in what file it should be located while it refers to several aspects of the app. Eg: `show_camera_video()` was put in the `homepage.py` file for convenience but not sure it's the right place from a logical point of view.

### UX
Where could we put the video according to you? It should probably be displayed somewhere else but it's a good start to put in the left side bar for now. 

**See the demo bellow:**

Path Helper Demo


app/homepage.py
```diff
@@ -57,6 +57,60 @@ def choose_map_style(n_clicks):
...
+#This function returns the user selection area in the left side bar
+def user_selection_area():
+    """ This function returns the basic dash component of the left side bar
+    of the monitoring platform and has been created because it is used several time within the app.
+
+    Parameters:
+    None
+
+    Returns:
+    html dash components: A set of components that composed the left side bar of the app
```
nrslt, line 69: I know it does not add anything to your code but it's a 'list' of components and not a 'set' :)

app/main.py
```diff
@@ -81,22 +81,44 @@ def display_page(pathname):
 
 @app.callback(Output('alerts_info', 'children'),
               [Input('geojson_alerts', 'hover_feature'),
-              Input('markers', 'n_clicks'),
               Input('markers', 'hover_feature')])
```
nrslt, line 84: just a question here: I am not sure I understand where the 'hover_feature' come from for 'geojson_alerts' and 'markers', I have been searching in the code for the past 20 minutes and do not find it. Is it inherent to dl.Geojson layers and thus automatically generated? Thanks for the heads up :)

nrslt (commented):
Nice job @VincGargasson ! I just had a question concerning the 'hover_feature' in the main file callbacks since I'm not very familiar with it and I did not find the info in the code..
Apart from that, I am not chocked by the video function being placed inside homepage.py. It could also be in utils.py I guess?

app/main.py
nrslt: Just a remark, should we let this whole section inside its own branch while it is still in development (we do not have the photos or real videos) and merge it when it's complete ?

VincGargasson: > Apart from that, I am not chocked by the video function being placed inside homepage.py. It could also be in utils.py I guess?

Yes of course. But I struggled to import functions from `homepage.py` into the other files (don't know why). So I prefered to stay in this file.

> Is it inherent to dl.Geojson layers and thus automatically generated?

Yes. It's part of the ~10 available properties of this components. ;)

> Just a remark, should we let this whole section inside its own branch while it is still in development (we do not have the photos or real videos) and merge it when it's complete ?

I'm not very keen of this solution because if we take too much delay to bring our first video, it will take some time to deal with conflict between the branch and the master that might have changed a lot.
I can maybe do like Sebastien for the graphs and disable it until we have proper data?

@Akilditu @pechouc  what do you think?

Akilditu:
> We have quite a few function and It can be sometime complicated to know in what file it should be located while it refers to several aspects of the app. Eg: `show_camera_video()` was put in the `homepage.py` file for convenience but not sure it's the right place from a logical point of view.

I think that the original idea was to define dash-specific functions in related `alerts.py` and `risks.py` and the transversal ones in `utils.py` or so. As we're moving to a 'centralized approach' around homepage, I agree that we may externalize functions definitions.
Let's discuss this @ the next weekly.  

> ### UX
> Where could we put the video according to you? It should probably be displayed somewhere else but it's a good start to put in the left side bar for now.

It's a good thing to retreive it here for now, we could also wan this to be displayed below the map with more description about the altert, but this is a good first draft !
In addition to this we may have to deal with pictures (frames) of a fire detection more than videos. The inference is indeed done with frames for now.

Akilditu (approved):
Thanks for the PR @VincGargasson ! 

Everything seems fine to me, as mentioned in my comments the next feature improvement may take frames instead of proper videos.

app/main.py
Akilditu: Having an experiment branch could be an option here, let's discuss this tomorrow at our weekly meeting 👍

VincGargasson:
@Akilditu @nrslt  I brought the correction you mentionned and resolved the conflicts my PR had with @pechouc 's
Could you give me your stamp if it's good for you? :)

PR #18, opened 2020-11-11
Name & logo

nrslt:
- added a app.title in main.py
- created an assets/ directory that stores the favicon.ico file, necessary to generate a browser tab icon

Screenshot 2020-11-11 at 17 25 07


frgfm:
@nrslt Thanks for the PR! Would you mind sharing a screenshot of how it renders? (so that we can keep track of evolutions when scrolling through merged PRs)

app/main.py
```diff
@@ -31,6 +31,7 @@
 
 # We start by instantiating the app (NB: did not try to look for other stylesheets yet)
 app = dash.Dash(__name__, external_stylesheets=[dbc.themes.UNITED])
+app.title = 'PyroNear'
```
frgfm, line 34: "Pyronear - Monitoring platform" to be more specific perhaps?

frgfm (commented):
Thanks again! I added some comments. Also, could you check if you find the proper icons over here : https://github.com/pyronear/pyronear.github.io/tree/master/img 
If so, it would remove the need to add some content in `app/assets`!

app/main.py
nrslt: sure, your call!

nrslt:
I changed the .ico to put the right one and also changed the title to "Pyronear - Monitoring platform"
the screenshot is in the PR description above!

frgfm:
@nrslt Thanks a lot!
My bad my previous comment wasn't very clear: is it possible to refer to the distant favicon? rather than committing it here?

nrslt:
@frgfm from what I've read so far, it seems to be only possible to change the favicon by adding a .ico file in the 'assets' folder that is at the root of the app..
I've tried to find other solutions but at this point I do not know how to use a distant url to specify the favicon file!

frgfm:
Thanks for investigating @nrslt ! Let's merge the PR then :ok_hand:

frgfm (approved):
Thanks a lot!

nrslt:
great can you merge it, then? :)

frgfm:
Yes sure! fyi, when you're the author of the PR and you have writing access to the repo, it's better if you do it :) (in case things go south, you're the most appropriate person to figure out what could have happened)

PR #23, opened 2020-11-30
feat : Added new markers types and live alerts features

### LINKED ISSUE #14: Adding Popup and Camera Video placeholder
## **Scope of the issue**

In order to display more informations to our future user, we need to have placeholder to display camera information.

**Suggestions**

Camera specs could be displayed in a popup
Video could be displayed in the side bar for now. Later this could be shown in a Popup maybe or in a dynamic SideBar.

Akilditu:
This PR aims at adding new components, methods and callbacks to handle live alerts and markers uses better.

Fixes #14 

Here is a snapshot of how the whole app renders depending on both `no alert` and `alert` states that can be set by radio items:

demo_platform

Few details because discussed a lot before among issues and meetings but I remain of course available for any matter.


app/alerts.py
```diff
@@ -19,25 +19,28 @@
...
 # Map layer
 # The following block is used to determine what layer we use for the map and enable the user to change it
 
 # This function creates the button that allows users to change the map layer (satellite or topographic)
 def build_layer_style_button():
 
     button = html.Button(children='Activer la vue satellite',
-                         id='layer_style_button')
+                         id='layer_style_button',
+                         className="btn btn-warning")
```
nrslt, line 43: c'est la première fois que l'on ajoute ce `className`, non ? quel est l'intérêt ?

app/alerts.py
```diff
@@ -88,50 +91,108 @@ def build_alerts_geojson():
...
+            children="Départ de feu, cliquez-ici !",
+            color="danger",
+            block=True,
+            id='alert_button'
+        )
+    else:
+        alert_button = ""
+
+    # Building alerts_markers objects and wraps them in a dl.LayerGroup object
+    if value == 0:
```
nrslt, line 151: c'est peut-être un détail je ne sais pas mais c'est pas mieux de grouper ces deux if/else en une seule ?
if value==0: on instancie le alert_button et le marker ?
else on fait rien

app/alerts.py
```diff
@@ -88,50 +91,108 @@ def build_alerts_geojson():
...
-# Once we have the positions of cameras, we output another GeoJSON object gathering these locations
-def build_alerts_markers():
-    markers = dl.GeoJSON(data=get_camera_positions(),
-                         id='markers')
+    return alert_button, alerts_markers_layer
+
+
+# This function either triggers a zoom towards the alert point each time the alert button is clicked
+# and sets the default zoom and center params for map_object
+def define_map_zoom_center(n_clicks):
```
nrslt, line 180: une proposition de nom: zoom_on_alert() ou bien alert_zoom(), c'est plus explicite ou pas ?

app/homepage.py
```diff
@@ -23,16 +23,42 @@
...
+from utils import build_live_alerts_metadata
+
 
 # ------------------------------------------------------------------------------
 # Before moving to the app layout
 
+#Fetching reusable alert metadata
+alert_metadata = build_live_alerts_metadata()
+if alert_metadata:
+    frame_url = alert_metadata["media_url"]
```
nrslt, line 36: FRAME_URL: quand on définit une variable globale, on met en maj ?

app/homepage.py
```diff
@@ -23,16 +23,42 @@
...
+
+
+# This function creates a radio button in order to simulate an alert event that will be later catched through the API
```
nrslt, line 39: caught ;)

app/homepage.py
```diff
@@ -57,6 +83,63 @@ def choose_map_style(n_clicks):
...
+            html.P(id="hp_alert_frame_metadata")                         # Displays alert_frame and alert_metadata
+            ]
+
+
+# Displays alert_frame and metadata related to a specific alert_markers after a click on display_alert_frame_btn
+def display_alerts_frames(feature=None):
+
+    # Fetching alert status and reusable metadata
+    alert_metadata = build_live_alerts_metadata()
+    alert_lat = str(alert_metadata["lat"])
```
nrslt, line 105: est-ce nécessaire de passer en str() les trois variables si tu les utilises dans une f-string ensuite ?

app/alerts.py
Akilditu: @nrslt thanks for reviewing, I'd rather that we keep on commenting in english for unity purposes and foreign contributions that may occur 👍

This className allow to custom the button color while using  html.Button as illustrated [here](https://getbootstrap.com/docs/4.0/components/buttons/)

app/alerts.py
Akilditu: Yep, indeed makes more sense !

app/alerts.py
Akilditu: I think that we should keep this 'define', 'build', 'display' logic to maintain readability and unity among methods. I also find it suitable to mention `zoom` and `center` : the 2 params updated here

nrslt (approved):
Nice job @Akilditu, thanks!
I just added some comments on the form rather than the content. This is a really cool intermediate stage from which we'll be able to iterate and implement some of our UX specialists' recommendations :)

My only question: do we know how the `alert_radio_button` will be linked to the API?

Moreover, at this point, the `build_live_alerts_metadata()` function stores a single alert information, I guess we'll have to take into account the 'multiple alerts at once' scenario, though we can deal with it later.

app/homepage.py
Akilditu: Should not have let this `if alert_metadata:`, thanks for pointing this out

app/homepage.py
Akilditu: Yep, good point. I was stucked trying to use `dcc.Markdown` that's why I tried to str them before giving up with this good old format 😄

app/alerts.py
nrslt: yup, I hadn't read the callback yet so it makes more sense as you say to keep the `zoom` and `center` reference in the function name

VincGargasson (approved):
@Akilditu  thanks for this great job! 
A big part of those changes has been reviewed by Nico and I. The pieces you added looks consistent with what we did. 
Remaining action will be to replace build_live_alerts_metadata() by real API data and store it in cache.

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
-        popup = ["Ville: {} <br>\
-                 Connexion Alerte CODIS: {} <br>\
-                 Nombres Devices: {}".format(area, alert_codis, nb_device)]
-        markers.append(dict(lat=lat,
-                            lon=lon,
-                            area=area,
-                            alert_codis=alert_codis,
-                            nb_device=nb_device,
-                            popup=popup))
+        markers.append(dl.Marker(id=f'site_{i}',    # Necessary to set an id for each marker to reteive callbacks
```
pechouc, line 113:
```suggestion
        markers.append(dl.Marker(id=f'site_{i}',    # Necessary to set an id for each marker to receive callbacks
```

Small typo in the comment. 

Just to confirm and if I am not mistaken, the `site` table that we will call via the API contains a `site_id` which we will be able to use instead of the index of the for loop here, right?

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+                                 position=(lat, lon),
```
pechouc, line 114: Probably not a priority for the moment but I wonder whether it is a good practice to define an extensive set of variables which we use only once, a few lines below (like `lat`, `lon`, `site_name`...). But clearly, we can move forward with this until demo day, as variable names are explicit.

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+                                 children=[dl.Tooltip(site_name),
+                                           dl.Popup([html.H2(f'Site {site_name}'),
+                                                     html.P(f'Coordonnées : ({lat}, {lon})'),
+                                                     html.P(f'Nombre de caméras : {nb_device}')])]))
+
+    # We group all dl.Marker objects in a dl.LayerGroup object
```
pechouc, line 120:
```suggestion
    # We group all dl.Marker objects in a dl.MarkerClusterGroup object
```

I guess this has changed since the comment was written down!

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+    markers_layer = dl.MarkerClusterGroup(children=markers, id='sites_markers')
```
pechouc, line 121: Maybe we can call it `markers_cluster` instead of `markers_layer` as the type of object seems to have changed?

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+def build_alerts_elements(value):
+
+    # Fetching alert status and reusable metadata
+    alert_metadata = build_live_alerts_metadata()
+    alert_lat = alert_metadata["lat"]
+    alert_lon = alert_metadata["lon"]
+    alert_id = str(alert_metadata["id"])
+
+    if value == 0:
+        # Building the button that allows users to zoom towards the alert marker
```
pechouc, line 140: Just an idea for later (I am fully comfortable with the idea of pushing the code in that form), but the clickable banner might be problematic if we have more than one alert on the map at some point. On which marker will it zoom?

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+        alerts_markers = [dl.Marker(
+            id="marker_{}".format(alert_id),   # Setting a unique id for each alerts_markers
+            position=(alert_lat, alert_lon),
+            icon=icon,
+            children=[dl.Popup(
+                [
+                    html.H2("Alerte détectée"),
+                    html.P("Coordonées : {}, {} ".format(alert_lat, alert_lon)),
+                    html.Button("Afficher les données de détection",
+                                id=("display_alert_frame_btn{}".format(alert_id)),  # Setting a unique btn id
```
pechouc, line 163:
```suggestion
                                id=("display_alert_frame_btn_{}".format(alert_id)),  # Setting a unique btn id
```

Amazingly low value-added comment 😅

But so far, for sites and alert markers, we always add an underscore before the id.

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
+        )
+        # Building alerts_markers objects and wraps them in a dl.LayerGroup object
+        icon = {
+            "iconUrl": 'https://marsfireengineers.com/assets/images/resources/firedetection.png',
+            "iconSize": [40, 40],       # Size of the icon
+            "iconAnchor": [20, 20]      # Point of the icon which will correspond to marker's location
+            # "popupAnchor": [-3, -76]  # Point from which the popup should open relative to the iconAnchor
+        }
+        alerts_markers = [dl.Marker(
+            id="marker_{}".format(alert_id),   # Setting a unique id for each alerts_markers
```
pechouc, line 155:
```suggestion
            id="alert_marker_{}".format(alert_id),   # Setting a unique id for each alerts_markers
```

To make it more explicit in case we want to call that id in a call-back at some point?

app/alerts.py
```diff
@@ -88,50 +91,105 @@ def build_alerts_geojson():
...
-    markers = dl.GeoJSON(data=get_camera_positions(),
-                         id='markers')
+    # Fetching alert status and reusable metadata
+    alert_metadata = build_live_alerts_metadata()
+    alert_lat = alert_metadata["lat"]
+    alert_lon = alert_metadata["lon"]
+
+    # Defining center and zoom parameters for map_object
+    if n_clicks > 0:
+        center = [alert_lat, alert_lon]
```
pechouc, line 186: I really like the fact that everything is already variabilized outside of the `build_live_alerts_metadata` function 😄

app/homepage.py
```diff
@@ -57,6 +82,63 @@ def choose_map_style(n_clicks):
...
+            html.P(id="hp_alert_frame_metadata")                         # Displays alert_frame and alert_metadata
+            ]
+
+
+# Displays alert_frame and metadata related to a specific alert_markers after a click on display_alert_frame_btn
+def display_alerts_frames(feature=None):
+
+    # Fetching alert status and reusable metadata
+    alert_metadata = build_live_alerts_metadata()
+    alert_lat = (alert_metadata["lat"])
```
pechouc, line 104: Is there any use to the parentheses that frame `alert_metadata["lat"]`? 🧐

app/homepage.py
```diff
@@ -80,52 +162,57 @@ def meteo_graphs(display=False):
 # Content and App layout
 # The following function is used in the main.py file to instantiate the layout of the homepage
 
+# Body container
```
pechouc, line 165: Isn't this part a duplicate of what is inside the `Homepage` function? Sorry if I am missing something obvious, but is there any reason we should keep it?

app/main.py
```diff
@@ -80,36 +79,49 @@ def display_page(pathname):
 # ------------------------------------------------------------------------------
 # Callbacks related to the "Alertes et Infrastructures" dashboard
 
+# Fetching reusable alert metadata
+alert_metadata = build_live_alerts_metadata()
```
pechouc, line 83: Isn't this contradictory with the fact that we are calling the `build_live_alerts_metadata` function in each of the functions that are related to alerts in the other files (notably `build_alerts_elements` or `build_alerts_markers` in `alerts.py` and `display_alerts_frames` in `homepage.py`)? Shouldn't these functions take an `alert_metadata` as argument so that we don't need to create it  each time we call these functions?

pechouc (commented):
Thanks a lot for this great PR @Akilditu! 

I have left a number of comments but these are either typos / very small suggestions or more long-term issues with which we probably don't want to deal at the moment. 

The only question that might be somewhat substantial is related to the loading of `alert_metadata`: in `main.py`, we make a first call at the `build_live_alerts_metadata` function, so that we can load the data only once. But in other files, all functions related to alerts also call the `build_live_alerts_metadata` function on their own if I am not mistaken. Shouldn't they instead take the `alert_metadata` object that has been instantiated as an argument? 

**EDIT:** Oops, I had not seen Vincent's comment above mine, especially the part about `alert_metadata` and storing it in cache; feel free to ignore my remark if I have just missed the discussion on this topic.

app/alerts.py
nrslt: yes I think you're right @pechouc, the f-string id was the previous way of getting a marker, from now on we should have a real site_id

Akilditu:
> Remaining action will be to replace build_live_alerts_metadata() by real API data and store it in cache.

Indeed this is the next PR purpose : fetching API data for real !

Akilditu:
> My only question: do we know how the `alert_radio_button` will be linked to the API?

It will be deleted and the function build_alerts_elements() will be triggered by the API response

> Moreover, at this point, the `build_live_alerts_metadata()` function stores a single alert information, I guess we'll have to take into account the 'multiple alerts at once' scenario, though we can deal with it later.

I was thinking about iterating through the API response DataFrame for that prupose

app/homepage.py
Akilditu: Not sure ... 😅

Akilditu:
> The only question that might be somewhat substantial is related to the loading of `alert_metadata`: in `main.py`, we make a first call at the `build_live_alerts_metadata` function, so that we can load the data only once. But in other files, all functions related to alerts also call the `build_live_alerts_metadata` function on their own if I am not mistaken. Shouldn't they instead take the `alert_metadata` object that has been instantiated as an argument?

>Isn't this contradictory with the fact that we are calling the build_live_alerts_metadata function in each of the functions that are related to alerts in the other files (notably build_alerts_elements or build_alerts_markers in alerts.py and display_alerts_frames in homepage.py)? Shouldn't these functions take an alert_metadata as argument so that we don't need to create it each time we call these functions?

You mean smth like this : `def build_alerts_elements(value, alert_metadata=build_live_alerts_metadata()):` ?

nrslt (approved):
TOP !

PR #28, opened 2020-12-09
Markers customization

nrslt:
Hi guys,
I just wanted to quickly add the new markers to this existing PR.

- blue one with tower icon = site marker
- red one with Pyronear icon = alert marker
- grey one with Pyronear icon = old fire marker

at this point it looks like this:
Screenshot 2020-12-09 at 12 20 51

To improve these, we will need to:
- rethink clearly the way we want to display alerts + sites,
- monitor the iconAnchor parameter so that the tip of the icon is correctly placed on the right coordinates (still unclear how this parameter behaves),
- change the iconURL path


Akilditu:
> * iconURL path

Thanks a lot for this PR, indeed this would be a good idea to host those icons somewhere else that in assets folder. 
@frgfm maybe as we did for logos under pyronear.org ?

Akilditu (changes requested):
Thanks a lot for this PR !

Could you please resolve the 2 following flake8 errors : 

```
./app/alerts.py:114:55: E261 at least two spaces before inline comment
./app/alerts.py:155:56: E261 at least two spaces before inline comment
```

frgfm:
@Akilditu yup, I think that would be the best thing to do :ok_hand: 
We can also make a mock release on the website repo, and put all logos in the attachments! Whichever you think is best

frgfm:
@nrslt if the reason the PR was closed is only about updating the branch, you can merge master (or another branch for that matter) ;) 
Would be better than reopening the same PR

PR #29, opened 2020-12-10
feat: Added support of API interaction within the web-app

### LINKED ISSUE #19: [API] Enabling user access and data inputs from pyro-api
Hi API team 👋 

The 2 following tasks could be nice first use-cases to enable pyro-api calls from pyro-platform : 

1. **User Access** : being able to request the `user` route to grant an access to the platform --> tests with pyro-platform dev team could be a nice first step and then retrieve the user `name` as the below 'SDIS Pyrodev' header 

Capture d’écran 2020-11-18 à 18 33 02

2. **Installations positions** : being able to request the `installation` route to replace the `app/data/cameras.csv` and then create the `camera_positions` DataFrame --> 

https://github.com/pyronear/pyro-platform/blob/c6a5aa619ecbda619f1cf5c1cf654f9c912d6d13/app/alerts.py#L102
cc @frgfm, @fe51, @jeanpasquier75, @florianriche

Thanks

frgfm:
Hey there :wave: 

As discussed earlier with @pechouc, we figured you might need to be able to access API features later. This PR does that exactly, now in your `main` or any other module, you can do the following:
```
from services import api_client
# then you can use any method of the client like so 
# api_client.blabla_my_method()
```

The supported methods of the API client are available in its documentation over here: https://pyronear.org/pyro-api/

Small details about authentication:
1. Currently I set them as environment variables, check the readme (suggestion: create a .env file with those variables and use the docker option to be able to test it just like in production).
2. On the long-term, you could use that as an authentication mechanism --> when the user accesses the platform, she/he enters some creds that are passed down to instantiate the `api_client` (any wrong creds will throw an error)

Successfully merging this PR will close #19 

Any feedback is welcome!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
+        env:
+          API_URL: ${{ secrets.API_URL }}
+          API_LOGIN: ${{ secrets.API_LOGIN }}
+          API_PWD: ${{ secrets.API_PWD }}
         run: docker-compose up -d --build
       - name: Ping app inside the container
         run: nc -vz localhost 8050
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,13 @@ ENV PYTHONUNBUFFERED 1
 COPY ./requirements.txt /usr/src/app/requirements.txt
 
 # install dependencies
-RUN pip install --upgrade pip setuptools wheel \
+RUN apt-get update \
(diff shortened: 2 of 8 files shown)


Akilditu:
> 2\. enters some creds that are passed down to instantiate the `api_client` (any wrong creds will throw an error)

Yes this could be a solution but we also could also imagine enhancing the user route for user access purpose

Akilditu (commented):
Thanks for this PR!

I was actually doing a mecanism using a config.py with auth and creds and then git igoring it.

But I guess that your option is smoother.

Akilditu (approved):

PR #30, opened 2020-12-11
fix: Fixed non-docker running of the Flask server

frgfm:
In #29, the need to have env variables was introduced. This PR introduces the following features:
- fixes env variables loading when not using Docker
- improves existing CI workflows

Any feedback is welcome!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,11 +2,9 @@ wheel>=0.29
 pandas>=0.24.2
 numpy>=1.18.4
 dash>=1.16.2
-dash-core-components>=1.12.1
-dash-html-components>=1.1.1
 dash-bootstrap-components>=0.10.6
 dash-leaflet>=0.1.4
-plotly>=4.7.1
 gunicorn>=20.0.4
 Flask-Caching>=1.3.3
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -2,11 +2,9 @@ wheel>=0.29
 pandas>=0.24.2
 numpy>=1.18.4
 dash>=1.16.2
-dash-core-components>=1.12.1
-dash-html-components>=1.1.1
```
pechouc, line 6: I am not sure to understand why we are erasing these requirements here?

pechouc (approved):
Thanks a lot @frgfm! I am not an expert at all and I don't really have enough time to dig deeper but I fully trust your input 😄

I just have a small question about requirements that are displayed as erased in the `requirements.txt` and which we are using for the web-app. Sorry if I have missed something obvious about this!

requirements.txt
frgfm: This is quite optional, it's more cosmetics than anything else: they are direct dependencies of  `dash` so installing dash installs these :)

pechouc: Oh ok, I thought I had to install them separately when I first installed dash, nice catch!

PR #31, opened 2020-12-12
feat: fetching alert data from pyroclient api

Akilditu:
This PR aims at:

- Fetching pyroclient api data mainly to retrieve detection`frame_url`
- Introducing the `dcc.interval` dash component that acts as a timer exec api fetches every 10 sec
- Disabling the `radio_items` that used to manually trigger alerts. Those could still be enabled for debug purposes

Notes:

- For now, once an alert is fetched and the `display_alert_frame_btn` is clicked, the intervall component (timer) is disabled untll webapp reload. This mecanism has to change in a next milestone

Thanks for reviewing!


app/main.py
```diff
@@ -254,14 +247,78 @@ def change_zoom_center(n_clicks=None):
...
 
-    return build_alerts_elements(value, alert_metadata)
+
+@app.callback(
+    Output('interval-component', 'disabled'),
+    [Input("display_alert_frame_btn{}".format(alert_id), 'n_clicks')])
+def callback_func_start_stop_interval(n_clicks):
+    '''
+    This one detects the number of clicks the user made on a display_alert_frame_btn.
+    If 1 click is made, the function disables the interval component.
```
pechouc, line 316: I am not sure to understand the rationale behind this callback: if we stop refreshing the data fetched by the web-app as soon as the user has clicked on an alert, he might miss detections that are made afterwards, no? Or maybe this is only for the demo-day to ensure that the information about the alert sent by Matteo and the associated image, etc are stable?

app/homepage.py
```diff
@@ -26,24 +26,20 @@
...
 # Before moving to the app layout
 
-#Fetching reusable alert metadata
-alert_metadata = build_live_alerts_metadata()
-frame_url = alert_metadata["media_url"]
-
 
-# This function creates a radio button in order to simulate an alert event that will be later caught by the API
+# This function creates a radio button for debug purposes, it simulates an alert event that will be
+# ater caught by the API
```
pechouc, line 34: Small typo in the comment and maybe we don't need the "that will be later caught by the API" anymore?

pechouc (commented):
Seems to work just fine, thanks a lot 😄

I just added one very minor comment and a question about the new `callback_func_start_stop_interval` callback!

app/main.py
```diff
@@ -254,14 +247,78 @@ def change_zoom_center(n_clicks=None):
...
+    Input('interval-component', 'n_intervals'))
+def fetch_alert_status_metadata(n_intervals):
+    '''
+    This callback takes as input the interval-component which acts as a timer,
+    scheduling API metadata fetches and defining alert status
+    '''
+    # Fetching live alerts where is_acknowledged is False
+    response = api_client.get_ongoing_alerts().json()
+    all_alerts = pd.DataFrame(response)
+    live_alerts = all_alerts.loc[all_alerts['is_acknowledged'] == False]
```
pechouc, line 262:
```suggestion
    live_alerts = all_alerts.loc[~all_alerts['is_acknowledged']]
```

Flake8 doesn't like this `== False` apparently, I just proposed another version that should work the same and satisfy styling requirements.

app/main.py
```diff
@@ -254,14 +247,78 @@ def change_zoom_center(n_clicks=None):
...
+        alert_status = 0
+        img_url = ""
+        return build_alerts_elements(img_url, alert_status, alert_metadata)
+    else:
+        alert_status = 1
+        # Fetching last alert
+        last_alert = live_alerts.loc[live_alerts['id'].idxmax()]
+
+        # Fetching last alert frame url
+        img_url = api_client.get_media_url(last_alert['media_id']).json()["url"]
```
nrslt, line 275: So cool that we're able to get the image this way thanks to the API!!

app/main.py
nrslt: I'll be interested to get more insight on this topic, as I don't really get the goal with this callback

nrslt (dismissed):
Really cool work, thanks for the great PR @Akilditu ! I'm also wondering what is the idea behind the last callback, such as @pechouc, though everything seems perfect otherwise!

app/main.py
Akilditu: > Or maybe this is only for the demo-day to ensure that the information about the alert sent by Matteo and the associated image, etc are stable?

Exactly ! As mentioned in notes, this is a workaround and should be changed in the future

app/homepage.py
Akilditu: Done

app/main.py
Akilditu: Thanks, done

app/main.py
pechouc: 🤦‍♂️ I should have read more closely, sorry about that!

pechouc (approved):
Thanks a lot for the edits!

PR #32, opened 2020-12-12
fix: Fixed gunicorn process on Heroku

frgfm:
Hey there :wave: 

Some of you noticed that the Heroku app has been crashing recently. So I investigated and fixed the problem:
- first of all, the crash is not due to our recent commits (most likely, one of the dependencies released a new version and we only have lower bound constraints)
- while for dev, we run the server either with python or docker (which is verified in the unittest), the Heroku app is run using the [Procfile](https://github.com/pyronear/pyro-platform/blob/master/Procfile). The corresponding command runs the server differently in a subprocess (using [gunicorn](https://gunicorn.org/) for the curious ones)
- so the actual command to run the production server was not tested, I added this in the CI workflow
- now, to fix the issue, the production app is run looking for the `server` attribute of the `app` object in the main module. A few commits ago, this was working by itself, so that's why I suspect a change in dependency versions. Anyway, I fixed the issue by forcing the creation of a `server` attribute of the `app` object

Any feedback is welcome!


Akilditu (dismissed):
Thanks a lot for investigating and resolving this !

PR #33, opened 2020-12-13
Alert acknowledgement

pechouc:
# Goal of the PR 

- Main goal of this PR is to add the checkbox that allows the user to acknowledge an alert when he has clicked on the marker and looks at the associated popup. 

- This PR also brings two small fixes:
  - Rewording of historic fire markers;
  - Changing the required version of `dash-leaflet` from ">=0.1.4" to "==0.1.4". My guess is that the Heroku machine was pip installing the "0.1.12" version which does not support the `dlx.choropleth` object anymore, so that the "Niveaux de risque" map could not be displayed.

# Main changes 

- Addition of the checkbox object in the popup of the alert marker 

- Creation of the callback that erases the checkbox once it is checked and changes the related message

- Modification of the last callback that disables the `dcc.Interval()` component
  - The interval component is now disabled as soon as the user clicks on the alert marker 
  - Goal is to avoid the alert to be reloaded after the user has checked the acknowledgement checkbox 

# Screenshots 

1. The user clicks on the alert marker (NB: interval component is disabled from this point on)

Capture d’écran 2020-12-13 à 09 35 55

2. The user checks the box to acknowledge the reception of the alert

Capture d’écran 2020-12-13 à 09 36 06

3. The user clicks on other buttons (any button on the "Alertes" map) 

Capture d’écran 2020-12-13 à 09 36 16

4. The user can click again on the alert marker and the checkbox is not accessible anymore

Capture d’écran 2020-12-13 à 09 36 19

# Question 

With this preliminary solution, if the user acknowledges the alert, moves to the "Niveaux de Risques" dashboard and comes back to the "Alertes" map, the alert data is reloaded and the acknowledgement checkbox can again be checked. I think it is because we are re-instantiating the interval component as part of the "Alertes" map.

I guess we will be able to bring a more solid solution to this issue once we can add an `if` statement in the callback that verifies whether the `is_acknowledge` field has already been changed to True or not. But at the moment, I am not sure to see a good solution to this problem without making the API call, etc. 

Do we agree to move forward with this imperfect solution for the demo-day and come back to it with a proper use of the client in the callback later on? 

cc @Akilditu @frgfm

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -163,6 +163,20 @@ def region_click_alerts(feature, radio_button_value):
         else:
             return None
 
+@app.callback(Output('acknowledge_alert_div_{}'.format(alert_id), 'children'),
+              [Input('acknowledge_alert_checkbox_{}'.format(alert_id), 'checked')])
+def acknowledge_alert(checkbox_checked):
+    if not checkbox_checked:
+        return [dbc.FormGroup([dbc.Checkbox(id='acknowledge_alert_checkbox_{}'.format(alert_id),
+                                            className="form-check-input"),
+                               dbc.Label("Confirmer la prise en compte de l'alerte",
+                                         html_for='acknowledge_alert_checkbox_{}'.format(alert_id),
+                                         className="form-check-label")],
+                              check=True,
+                              inline=True)]
+    elif checkbox_checked:
+        return [html.P("Prise en compte de l'alerte confirmée")]
+
 # ------------------------------------------------------------------------------
 # Callbacks related to the "Niveau de Risque" page
 
--- a/app/main.py
+++ b/app/main.py
@@ -325,10 +325,10 @@ def display_alert_frame_metadata(n_clicks_marker, img_url):
 
 @app.callback(
     Output('interval-component', 'disabled'),
-    [Input("display_alert_frame_btn{}".format(alert_id), 'n_clicks')])
+    [Input("alert_marker_{}".format(alert_id), 'n_clicks')])
 def callback_func_start_stop_interval(n_clicks):
     '''
-    This one detects the number of clicks the user made on a display_alert_frame_btn.
+    This one detects the number of clicks the user made on an alert marker.
     If 1 click is made, the function disables the interval component.
     '''
     if n_clicks is not None and n_clicks > 0:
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -311,10 +326,10 @@ def display_alert_frame_metadata(n_clicks_marker, img_url):
 
 @app.callback(
     Output('interval-component', 'disabled'),
-    [Input("display_alert_frame_btn{}".format(alert_id), 'n_clicks')])
+    [Input("alert_marker_{}".format(alert_id), 'n_clicks')])
```
Akilditu, line 329: So you're coming down to the `alert_marker` level because not possible to combine `"display_alert_frame_btn{}".format(alert_id), 'n_clicks'` and `'acknowledge_alert_checkbox_{}'.format(alert_id), 'checked'` as inputs as they don't occur as the same time right ?

requirements.txt
```diff
@@ -3,7 +3,7 @@ pandas>=0.24.2
 numpy>=1.18.4
 dash>=1.16.2
 dash-bootstrap-components>=0.10.6
-dash-leaflet>=0.1.4
```
Akilditu, line 6: Was it the reason why risk map was 'disabled' ?

Akilditu (approved):
Thanks a lot for this PR!

It seems that for now we'll deal with the interval component this way but we'll have a to find a proper solution later on as this is getting even more straightforward

app/main.py
pechouc: To be fully transparent, I came back to the marker level because this was the quickest fix I saw! 

But to answer your remark, for the reason you mention, we could not combine both inputs indeed. Maybe there is a syntax to have the callback triggered by one input **or** the other. Or maybe we could write the same logic with a `State`. 

But since this callback is not supposed to be kept anyway, I have not spent much time looking for other options at hand.

requirements.txt
pechouc: It's difficult to say certainly because the callbacks do not output error messages outside of the debugging/development mode of Dash. But I could recreate the issue locally by uninstalling `dash-leaflet` and running `pip install -r requirements.txt`. When coming back to the 0.1.4 version, the problem was solved: so, my guess is that this should work better with a `==`!

PR #34, opened 2020-12-13
Using real calculated risk scores for each department to be displayed on the map

nrslt:
Hi guys,

This PR adds real risk scores data to the web app.

At this point, the data is retrieved in a 'static' way from here 'https://github.com/pyronear/pyro-risks/releases/download/v0.1.0-data/pyrorisk_20200901.json'

Note that 3 departments do not have risk scores, though it won't be a problem for the presentation as they're all located in the Ile de France region.

When I'll have time I'll look for a better way to store the data in a dcc.Store() object while launching the app, then using it to 'populate' the risk map.

ezgif-4-1014a5943b3a

Tell me if you have any remark, and thanks @frgfm for the help!

@frgfm @Akilditu


app/config.py
```diff
@@ -8,3 +8,4 @@
 API_URL: str = os.getenv('API_URL')
 API_LOGIN: str = os.getenv('API_LOGIN')
 API_PWD: str = os.getenv('API_PWD')
+PYRORISK_FALLBACK: str = "https://github.com/pyronear/pyro-risks/releases/download/v0.1.0-data/pyrorisk_20200901.json"
```
frgfm, line 11: EOF missing over here

app/risks.py
```diff
@@ -34,10 +36,18 @@
 with open(Path(__file__).parent.joinpath('data', 'departements.geojson'), 'rb') as response:
     departments = json.load(response)
 
+# We fetch the deparment risk score json and store it in the risk_json variable
+risk_json = requests.get(cfg.PYRORISK_FALLBACK).json()
```
frgfm, line 40: Could you add the actual live request as well and comment it?
So that the switch will be easy once we solved the inference issues on the pyrorisk side :pray:

frgfm (commented):
Thanks for the PR! Some comments about integration on my side, would you have a screenshot of how the risk computation renders?

app/risks.py
nrslt: yes, definitely, I thought I had let it but I forgot apparently!

app/config.py
nrslt: check

frgfm (approved):
Looks good to me for the request part! I think we need another reviewer for the dash part. Could you add a screenshot of the loading part you implemented please?

frgfm:
@nrslt also I'm a tiny bit suspicious of that failed test. Could you try to run:
```
gunicorn --workers 1 -b 0.0.0.0:8050 --chdir ./app main:server
```
on your branch please? If this fails, we need to fix it

app/risks.py
```diff
@@ -34,10 +36,19 @@
 with open(Path(__file__).parent.joinpath('data', 'departements.geojson'), 'rb') as response:
     departments = json.load(response)
 
+# We fetch the deparment risk score json and store it in the risk_json variable
+# risk_json = requests.get("http://pyro-risks.herokuapp.com/risk/fr/2020-09-01").json()
```
Akilditu, line 40: Is this line useful ? If yes in for which scenario ?

Akilditu (changes requested):
Thanks a lot for this PR!

Agreed with @frgfm, could you please add a screenshot of how it renders + check this gunicorn issue. Then we'll be good to merge !

app/risks.py
frgfm: It was my suggestion: this is the line to use once the timeout will be taken care of on the risk prediction server :ok_hand: 
But on 2nd thought, we can add this later on and avoid displaying the URL publicly! So perhaps we should remove this in the end

nrslt: OK I deleted the line, we'll add it later

nrslt:
Thanks @frgfm @Akilditu I added a gif of how the app renders.

@frgfm the gunicorn command works on my side, though there is still this docker-gunicorn failed test.. We can discuss it quickly today if you want.

Akilditu (approved):
Thanks for the corrections and details ! 

If ok with gunicorn let's merge 👍

Issue #35, opened 2020-12-24
[Content] Large content files should be removed from repo


frgfm:
Simply put, graphics assets could indeed be stored in the git history as they have a very limited size (event though we could avoid it). However, content files like the French geojson or wildfire history data should be moved to release attachments or on cloud storage to preserve git history sanity.

They could be downloaded when the app starts and cached.

- [x] cameras.csv (#41)
- [x] departments.geojson (#39)
- [ ] historic_fires.csv (#81)

pechouc:
Hello @frgfm! 

Sorry for answering this issue so late, we have indeed added to our to-do list the removal of local data files from the repository. These are: 

- `cameras.csv` - Everything is ready to tackle this one thanks to the `get_sites` function of the API client (thanks a lot for the work done on that side by the way!). We are going to implement it soon with @abdi-adel 😊

- `departments.geojson` - For this one, we cannot directly use the file available online since the [website](https://france-geojson.gregoiredavid.fr) was recently down for several weeks in a row according to @chloeskt. Apparently, in the pyro-risk repository, it was attached to a [release](https://github.com/pyronear/pyro-risks/releases). So, what do you think is best between fetching the file from there or having it released in the pyro-platform repository too? If the first option is fine, we can implement it very shortly. 

- `historic_fires.csv` - There might be a bit more material for discussion here but I must admit I have no idea what would be the best practice. 
  - On the one hand, we haven't planned to update this list of past fires anytime soon and therefore, the file is probably bound to remain static for a while (which would justify attaching it to a release I guess). 
  - But in a way, it would make sense to store it in the database because (i) it could be useful not only to the platform team and (ii) this would be the occasion to script the logic that allowed us to preprocess the raw data found on data.gouv, so that we would be able to refresh it every time the dataset is updated.

frgfm:
No worries, we all had our hands busy!

- `departments.geojson`: for now, yes it would be a good idea to fetch it from the release when you start the app (if it's not already downloaded)
- `historic_fires.csv`: this one will in the end be available through the API as well. The key here would perhaps be to add some previous wildfires thoroughly in the API DB

pechouc:
Thanks for your answer 🙏

I have made a quick test in a notebook to check whether we can fetch the `.geojson` file from the release and it works just fine. We will implement it very soon in the app itself. 

[...]

PR #39, opened 2021-02-15
Fetching GeoJSON file from Pyro-Risk release

pechouc:
Hello everyone!

Following #35, here is a very light PR to switch from reading the local GeoJSON file in the `data` folder to fetching the file online, from a [release](https://github.com/pyronear/pyro-risks/releases/tag/v0.1.0-data) in the Pyro-Risk repository. 

The first commit thus changes:
- the `alerts.py` file to (i) move the file reading procedure outside of the `build_departments_geojson` function, (ii) modify this process from a local reading to the online one and (iii) adapt imports and comments as needed;
- the `risks.py` file in which the file reading procedure was already outside of the function;
- the `config.py` file, so as to store the URL towards the release in a clean way. 

The second commit simply consists in eliminating the unnecessary local file. 

My only interrogation is the following. Because the file is the same for both alerts and risks views, we could theoretically fetch the file only once for both, which would be more efficient. But at the same time, in `risks.py`, we are also fetching the estimated risk scores from the data science team and use them to modify the `departments` variable. So, it seems that it would require some work to merge both requests to the online file and it feels as a bit of an overkill to me. 

What is your opinion on this? Should we absolutely read the file only once for both maps?


chloeskt (changes requested):
Hi @pechouc ! Thank you for this PR, seems great !! 

I've had a look at this issue given by [Codacy](https://app.codacy.com/gh/pyronear/pyro-platform/pullRequest?prid=6956940). This is that there are security issues with `urllib`, especially due to the permitted schemes ([SO thread](https://stackoverflow.com/questions/48779202/audit-url-open-for-permitted-schemes-allowing-use-of-file-or-custom-schemes)). I guess that the security risk is not that important in our case, but we should try to have as many best practices as possible. Maybe we should use another library such as `requests` which do not face the same issue ? or make sure that the transfer protocol is http or https if you want to use ` urllib` ?

pechouc:
Thanks for the heads-up, I had missed the Codacy check! I have moved to `requests` in a new commit, which presents the additional advantage of harmonising these few lines with the process already used for getting the risk score.

chloeskt (approved):
That is great !

frgfm:
Just updated @chloeskt access to the repo, you now have write access :ok_hand: 
So @pechouc, you can move ahead with the merge if everything looks alright to you :)

chloeskt:
Thanks @frgfm ☺️

pechouc:
Wonderful, thanks a lot!

PR #40, opened 2021-02-18
chore: Updated license from GPLv3 to Apache 2.0

frgfm:
This PR introduces the following modifications:
- updated the license content
- updated license reference in the README
- updated all file headers
- added a unittest checking all python file headers

Any feedback is welcome!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,6 +48,19 @@ jobs:
           flake8 --version
           flake8 ./
 
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


fe51 (approved):

Akilditu:
Hi @frgfm thanks for the PR!

Maybe highlighting here the reasons motivating the license switch would be a good thing.

frgfm:
@Akilditu good idea: we reserve GPL-based licenses for repos that let available assets that required resources from contributors (training a machine learning model, labelling a dataset, etc.). The GPL license is the higher-end of open source licenses, while the Apache 2.0 is more appropriate for repos using existing open source frameworks.

PR #42, opened 2021-03-26
Alert screen for CODIS operational room

chloeskt:
### Motivation

Hi all ! This PR aims at transforming our platform into a multiple page app:
- the classic homepage that we already have
- the alert page: it is meant to be displayed on a big screen in CODIS's room. There is a sober screen when there are no alert and it switches to an alert mode when there are alerts.

As discussed, only the lastest event (as latest fire) is displayed on the screen. When multiple alerts are given by the RPI for the same event (same fire), we display a GIF of all images incoming (doubt relief for the firefighters). 

### Features

Multiple features have been added:
- no alert screen
- alert screen
- button on the navbar (top right corner) to access the alert page ("ecran de monitoring")
- css styles (to make an element blink)
- changes in rough layout

##### Changes in rough layout:

- Store `div_live_alerts_data`: in order to store the table of live alerts we get once the call to the API is done
- Store `last_displayed_event_id`: to keep track of the last event (fire) which has been displayed on the alert screen (useful for the GIF part)
- Store `images_url_current_alert`: this is a Dict[str, List[str]]. Keys: str of the event id. Values: List of all the urls which belong to this event id (again for GIF part)

This might change with the use of a webhook, but we will still need to store the data somewhere to make it accessible to multiple callbacks.

### Limits

As we know, the webhook will change many things, therefore some developments made here might become outdated. Also we will need to think about deleting some info at some point, especially for the Store `images_url_current_alert`, otherwise the dict will keep on growing. Once a event has been "acquitté", we can remove it from the dict. This will be easy but we need to wait for the webhook to get a better picture of the workflow needed. 

### Screens

When there are no alerts:

Capture d’écran 2021-03-28 à 14 15 48 (3)

When there is an alert:

Capture d’écran 2021-03-28 à 14 15 42 (3)


Akilditu:
Thanks for the PR, will review it asap 👍

app/main.py
```diff
@@ -68,11 +70,24 @@
...
+        html.Div(id="page-content", style={"height": "100%"}),
+        # Interval component to generate call to API every 10 seconds
+        dcc.Interval(
+            id="interval-component-homepage",
+            interval=10 * 1000
+        ),
+        # Hidden div to keep a record of live alerts data
+        dcc.Store(id="div_live_alerts_data", storage_type="memory"),
+        dcc.Store(id="last_displayed_event_id", storage_type="memory"),
+        dcc.Store(id="images_url_current_alert", storage_type="memory", data={}),
```
Akilditu, line 88: When the webhook will be ready I guess that all alerts metadata could be stored in one dcc.Store like `div_live_alerts_data` ?

app/main.py
```diff
@@ -92,6 +107,21 @@
...
+)
+def display_page(pathname):
+    """
+    This callback takes the url path as input and returns the corresponding page layout,
+    thanks to the instantiation functions built in the various .py files.
+    """
+    if pathname == "/alert_screen":
+        return alert_screen()
+    else:
+        return Homepage()
```
Akilditu, line 122: @pechouc we used to have this `display_page` callback, but I do not remember where it came and by what is has been replaced ?

app/alert_screen.py
```diff
@@ -0,0 +1,256 @@
...
+# The following block gathers elements defined above and returns them via the alert_screen function
+def alert_screen():
+    """
+    The following function is used in the main.py file to build the layout of the big screen page.
+    """
+    layout = html.Div(
+        children=[
+            dcc.Interval(
+                id="interval-component-alert-screen",
+                interval=3 * 1000
```
Akilditu, line 246: I'm not sure to understand why there are 2 `dcc.Interval ` here and inside `fire_images_div`, the `interval-component-img-refresh` would not be enough for the gif ?

Akilditu (commented):
Thanks a lot for this PR, overall great maybe adding 2 screenshots of how the 2 big screens renders would be nice :)

app/alert_screen.py
chloeskt: The Interval `interval-component-homepage` is used to call the API and update the alerts metadata (see function `update_alert_data()`). I could not use it to synchronize the callback `update_alert_screen()`: I had the error `a nonexistent object was used in an output of a dash callback` due to the fact that the div `core_layout_alert_screen` does not exist in the `homepage` (precisely because it is not meant to be on this page).

This is why I had to create the Interval `interval-component-alert-screen`: in order to update the alert page display (make the switch between the alert screen and the no alert screen when needed). 

And yes the Interval `interval-component-img-refresh` is used for the GIF.

app/main.py
chloeskt: It is mean to have multiple pages app. Here it means that `http://127.0.0.1:8050/alert_screen` will point to the `alert_screen()` function. Whereas `http://127.0.0.1:8050/` or `http://127.0.0.1:8050/whatever_you_want` will point to `homepage()`function

app/main.py
chloeskt: Yes ! But I think that we still need to have `last_displayed_event_id` (to keep track of what is the last event displayed on big screen) and `images_url_current_alert`(dict with event id and a list of images' urls). However for that last one, we will need to add a mechanism to erase past events (probably once the alert has be acquitted).

app/alert_screen.py
```diff
@@ -0,0 +1,256 @@
...
+        src=img_url,
+        style={
+            "position": "relative",
+            "width": "100%",
+            "height": "100%",
+            "object-fit": "contain",
+        },
+    )
+
+    # Fire alert image div (right of the screen)
```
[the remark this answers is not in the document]
chloeskt: Done

app/main.py
```diff
@@ -68,11 +70,24 @@
...
+app.layout = html.Div(
+    [
+        dcc.Location(id="url", refresh=False),
+        html.Div(id="page-content", style={"height": "100%"}),
+        # Interval component to generate call to API every 10 seconds
+        dcc.Interval(
+            id="interval-component-homepage",
+            interval=10 * 1000
+        ),
+        # Hidden div to keep a record of live alerts data
```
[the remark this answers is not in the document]
chloeskt: No, not really. According to the [documentation](https://dash.plotly.com/dash-core-components/store)

- The memory store reverts to the default on every page refresh
- The local store will take the initial data only the first time the page is loaded and keep it until it is cleared 
- The session store: same as the local store but will lose the data when the browser/tab closes 

Here I have the following ideas:
- for `div_live_alerts_data` : we get data from the API every 10 seconds. Therefore it's okay to loose it on refresh
- for `last_displayed_event_id`: cannot keep this variable in memory, it needs to be updated all the time as it is used in callbacks to know if we keep on displaying the same event or not
- for `images_url_current_alert`: I switched it to `session` because if the user refreshes the page (but let's hope he does not, don't think that Dash likes refresh), we will lose the urls. To avoid this, we need to store them into the browser.

app/main.py
Akilditu: I know and we used to have this multiple pages (/alerts, /risks) display page function. I was just wondering why we do not have it anymore but now I remember that deleted when we moved to a 'one pager' platform format.

chloeskt:
> Thanks a lot for this PR, overall great maybe adding 2 screenshots of how the 2 big screens renders would be nice :)

Thank you for your review ! I added two screenshots in the description of the PR 👌

app/main.py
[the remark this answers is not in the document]
Akilditu: Thanks for the detailed info.

> * for `images_url_current_alert`: I switched it to `session` because if the user refreshes the page (but let's hope he does not, don't think that Dash likes refresh), we will lose the urls. To avoid this, we need to store them into the browser.

Agreed on the `session` storage.

app/main.py
Akilditu: I'm not sure for `images_url_current_alert`, pretty sure that `div_live_alerts_data` can do the job but we'll discuss it later when having a better idea of how we'll interact with webhook data.

chloeskt: Ok sure, the webhook will change a bit the workflow anyways !

Akilditu (dismissed):
Thanks for changes and answers !

app/alert_screen.py
```diff
@@ -0,0 +1,247 @@
...
+# ----------------------------------------------------------------------------------------------------------------------
+# IMPORTS
+
+# Various modules provided by Dash to build the page layout
+import dash_bootstrap_components as dbc
+import dash_core_components as dcc
+import dash_html_components as html
+
+
+# Importing the pyro-API client
```
pechouc, line 24: High value-added comment incoming: I think we can get rid of this comment, right?

app/alert_screen.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+# ----------------------------------------------------------------------------------------------------------------------
+# CONTENT
+def no_alert_detected_screen():
```
pechouc, line 29: I am going to be a pain but do you think we could rename the function something like "build_no_alert_detected_screen" (or equivalent) to ensure consistency in the code?

app/alert_screen.py
```diff
@@ -0,0 +1,247 @@
...
+        "backgroundSize": "cover",
+        "position": "fixed",
+        "height": "100%",
+        "width": "100%",
+    }
+
+    return style
+
+
+def alert_detected_screen(img_url, alert_metadata, last_alert):
```
pechouc, line 47: Same comment as above, sorry 😬

app/alert_screen.py
```diff
@@ -0,0 +1,247 @@
...
+        "position": "fixed",
+    }
+
+    return layout_div, style
+
+
+# ----------------------------------------------------------------------------------------------------------------------
+# App layout
+# The following block gathers elements defined above and returns them via the alert_screen function
+def alert_screen():
```
pechouc, line 232: Maybe this one could be renamed into something like "AlertScreen" to keep the same naming convention as the one used for the "Homepage" function?

app/main.py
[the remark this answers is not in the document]
pechouc: Very clear, thanks a lot!

It will probably be worth testing this cache mechanism in depth after we have implemented the webhook-based alert flow, the login, etc. To investigate what happens when the page is refreshed, when the app is opened in two tabs simultaneously, etc. So that we can identify the behaviours we want to keep and those that we may want to eliminate.

app/main.py
pechouc: This was from this [beloved tutorial](https://towardsdatascience.com/create-a-multipage-dash-application-eceac464de91?gi=8a94d5ddbb31) but yes, we killed at some point during the DfG season.

app/main.py
```diff
@@ -427,15 +468,22 @@ def change_map_style_main(map_style_button_input, alert_button_input, map_style_
...
-     Output('live_alert_header_btn', 'children'),
-     Output('live_alerts_marker', 'children')],
-    Input('interval-component', 'n_intervals'),
-    State('map_style_button', 'children')
+    [
+        Output("img_url", "children"),
+        Output("live_alert_header_btn", "children"),
+        Output("live_alerts_marker", "children"),
+    ],
+    Input("interval-component-homepage", "n_intervals"),
```
pechouc, line 476: If I am not mistaken, this means that we have two callbacks (this one and the one right above) being triggered by the same input right? Couldn't they be merged into a unique callback then? 

Or maybe if we want to keep a notion of order in the triggering of the two callbacks, maybe this second callback could be triggered by changes to the alert data storage component? The input could directly be:

`Input('div_live_alerts_data', 'data')`.

app/main.py
```diff
@@ -546,11 +594,121 @@ def callback_func_start_stop_interval(n_clicks):
...
+            # new event, not been displayed yet
+            layout_div, style_to_display = alert_detected_screen(
+                img_url, alert_metadata, last_alert
+            )
+            return layout_div, style_to_display, last_event_id
+
+
+@app.callback(
+    Output("images_url_current_alert", "data"),
+    Input("interval-component-homepage", "n_intervals"),
```
pechouc, line 646: Again, I feel like this callback could be merged with one of the two callbacks triggered by `Input("interval-component-homepage", "n_intervals")` above. Typically, I think this one could be merged with the first one that updates the alert data storage component in the sense that we would centralise all API calls in the same callback. What do you think? 🙂

app/navbar.py
```diff
@@ -58,6 +58,26 @@ def Navbar(dropdown=False):
         className="mx-auto order-0",
         style={'color': 'white', 'align': 'center', 'justify': 'center'})
 
+    # Alert monitoring screen, to be displayed in CODIS
+    alert_screen = dbc.NavLink(
```
pechouc, line 62: Last low value-added comment I promise: what do you think of renaming this object "alert_screen_button" or something equivalent so that we don't get confused with the functions and component in the `alert_screen.py` file?

pechouc (commented):
Hello! 

Thanks a lot for this PR, it clarifies a lot of things and the alert screen seems to work perfectly 😊

I have added a few very dumb comments regarding naming / styling, I hope they are not too "relous". However, there is maybe a more substantial discussion regarding the three different callbacks being triggered by the same input (`Input("interval-component-homepage", "n_intervals")`). Let me know what you think about this!

app/alert_screen.py
chloeskt: Yes done

app/alert_screen.py
chloeskt: done

app/alert_screen.py
chloeskt: done

app/alert_screen.py
chloeskt: It depends. `Homepage` does not follow PEP8 convention (i.e. use of snake case), this is upper Camel case. Do we want to keep it that way ? I don't really care to be honest

app/main.py
chloeskt: I don't like to merge the two. They are doing very separate stuff: one is calling the API, the other is creating stuff on the map. The logic is very different. 

I'm proposing the following:
- `update_style_components_with_alert_metadata` with the input that you suggested

app/main.py
chloeskt: Again not sure because these two callbacks are doing very different things. I'd like to keep it so that one function = one task. And anyways I believe that the callback `update_alert_data` will be removed with the webhook

app/navbar.py
chloeskt: done !

chloeskt:
Thank you @pechouc for your review ! Mostly I agree with you (I even think that to improve the quality of the code, we should do some refactoring. But it is clearly not the priority). Regarding the merge of `update_alert_data` and `update_dict_of_images` I'm not sure. Also I think that many things will need to change once we get the webhook to work, we will probably need to discuss the workflow again !

chloeskt:
@Akilditu @pechouc I had to remove the following lines in `styles.css` due to issues reported by Codacy through `CSSLint` (which seems to be an unmaintained library with multiple issues ... StyleLint seems to be way better). 

```
@-ms-keyframes blink { 
  0% {
    opacity: 1;
  }

  50% {
    opacity: 0;
  }

  100% {
    opacity: 1;
  }
}
```

Many users had the same issue which has been reported to the creators of CSSLint, see [this issue on Github](https://github.com/CSSLint/csslint/issues/295). 

Let's hope that our users won't have IE < 10: the blink features won't work for them !

app/alert_screen.py
chloeskt: I've changed it to `AlertScreen`

pechouc: Thanks a lot! I agree that it is slightly unorthodox but I feel like using the styling convention of a class reflects well the "instantiation" aspect of this function, which does not really encapsulate a logic but gathers previously defined function to return an object that did not exist before. Maybe I'm going too far ahah and anyway, it was not very important but many thanks for the change 😊

pechouc (approved):
Thanks a lot for the changes, I agree with you: let's keep the callbacks as in the revised version! As you rightly mentioned, this logic will probably change significantly with the web-hook.

PR #43, opened 2021-03-30
Login

pechouc:
Hello everyone! 

This pull request introduces login in our app, which consists in a modal that is opened when the user arrives on the page and that contains a username/password form. The credentials are verified thanks to a login correspondences file, stored in the repository but encrypted with the [Simple Crypt](https://pypi.org/project/simple-crypt/) library. To open it, an environment variable `LOGIN_CORRESPONDENCES_KEY` is required, which corresponds to the password used for the encryption of the file (this allowed to add this variable in the secret configuration variables on Heroku). 

A new callback manages the login process, ie. checks the user's credentials and if these are validated, closes the modal. The callback also updates the `center` and `zoom` attributes of the map object, so that the user ends up with a view of his department only. It also fetches site devices data from the API which will probably prove useful later on, when dealing with the alerts flow. 

This summarises the main changes to `homepage.py` and `main.py`. As for `alerts.py`, I added a function which allows to build the vision angle of a camera based on various inputs (location of the device, yaw, opening angle, etc). 

**NB:** in `main.py`, I also commented the callback updating the zoom and center attributes of the map object based on clicks on the alert banner, to avoid any conflict with the login callback and because - if I am not mistaken - we are not keeping the banner in the next steps. 

Let me know if you have any comments or questions of course, here are a few screenshots of how it renders:

Capture d’écran 2021-03-30 à 09 35 10

Capture d’écran 2021-03-30 à 09 35 28

Capture d’écran 2021-03-30 à 09 35 23

Capture d’écran 2021-03-30 à 09 35 42


pechouc:
There seems to be an error with Docker when trying to install a dependency (`pycrypto`) of the Simple Crypt library. I will look into it as soon as possible!

requirements.txt
```diff
@@ -8,3 +8,5 @@ gunicorn>=20.0.4
 Flask-Caching>=1.3.3
 pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
 python-dotenv>=0.15.0
+pycryptodome>=3.10.1
+geopy>=2.1.0
```
chloeskt, line 12: I think that you forgot to add `simple-crypt>=4.1.7`

app/main.py
```diff
@@ -252,43 +400,43 @@ def acknowledge_alert(checkbox_checked):
...
-    [Output('map', 'center'),
-     Output('map', 'zoom')],
-    Input("alert_button_alerts", "n_clicks"),
-    State('map_style_button', 'children')
-)
-@cache.memoize()
-def change_zoom_center(n_clicks, map_style_button_label):
-    """
-    -- Zooming on the alert from the banner --
+# @app.callback(
```
chloeskt, line 403: Rather than commenting the code, I think we should erase it. Letting commented code is a bad practice I think, it makes the source code not really readable ...

app/homepage.py
```diff
@@ -287,6 +291,81 @@ def display_meteo_graphs(display=False):
...
+# Login modal
+# The following block defines the build_login_modal, which allows to instantiate the modal opened at the beginning of
+# each session for the user to enter his/her credentials.
+
+def build_login_modal():
+    """
+    This function simply creates and returns the login modal, which opens up at the beginning of each session to obtain
+    the credentials of the user.
+
+    Notice the backdrop='static' and keyboard=False arguments passed at the very bottom of the instantiation of the mo-
```
chloeskt, line 304: Thanks for explaining this !

app/homepage.py
```diff
@@ -48,6 +48,10 @@
 # ----------------------------------------------------------------------------------------------------------------------
 # CONTENT
 
+# Pyronear - Horizontal Logo
+pyro_logo = 'https://pyronear.org/img/logo_letters.png'
```
chloeskt, line 52: Shouldn't we put this in the `assets` folder rather ? I think that it doesn't fit well here

app/alerts.py
```diff
@@ -131,6 +153,52 @@ def build_sites_markers(dpt_code=None):
     return dl.MarkerClusterGroup(children=markers, id='sites_markers')
 
 
+def build_vision_polygon(site_lat, site_lon, yaw, opening_angle, dist_km):
+    """
+    This function allows to build the vision angle of a camera, ie. the zone covered by the detection device.
+
+    It takes as input:
+
+    - site_lat, the latitude of the detection device;
```
chloeskt, line 162: Nice function !! 

Quick question: will we have all these inputs through the API ?

app/alerts.py
```diff
@@ -72,6 +76,24 @@ def build_departments_geojson():
     return geojson
 
 
+# ----------------------------------------------------------------------------------------------------------------------
+# Site devices
+# The following block is used to load the data relative to site devices.
+# It is called in main.py by the callback that is itself triggered by the user's login.
+
+def get_site_devices_data(client):
```
chloeskt, line 84: Dumb comment: maybe we could remove the blank lines ?

app/main.py
```diff
@@ -81,6 +101,9 @@
         dcc.Store(id="store_live_alerts_data", storage_type="memory"),
         dcc.Store(id="last_displayed_event_id", storage_type="memory"),
         dcc.Store(id="images_url_current_alert", storage_type="session", data={}),
+
```
chloeskt, line 104: Other dumb comment: should we remove the blank line ?

app/main.py
```diff
@@ -173,6 +196,131 @@ def update_alert_data(interval):
...
+    [Output('login_modal', 'is_open'),
+     Output('form_feedback_area', 'children'),
+     Output('site_devices_data_storage', 'data'),
+     Output('map', 'center'),
+     Output('map', 'zoom')],
+    Input('send_form_button', 'n_clicks'),
+    [State('username_input', 'value'),
+     State('password_input', 'value')]
+)
+def manage_login_modal(n_clicks, username, password):
```
chloeskt, line 209: This is nicely explained, thank you !

chloeskt (changes requested):
Hi @pechouc nice PR, thank you very much. I've done some small comments. 

There is an issue with the callback `update_alert_data` as you can see on this screen. I've investigated a bit and its linked to the fact that now the `response = api_client.get_ongoing_alerts().json()` returns an empty json (and the corresponding dataframe has no columns ...). To prevent this issue, you can change the callback to: 

```
@app.callback(
    Output("store_live_alerts_data", "data"),
    Input('interval-component-homepage', 'n_intervals')
)
def update_alert_data(interval):
    """
    The following function is used to update the div containing live alert data from API.
    """
    # Fetching live alerts where is_acknowledged is False
    response = api_client.get_ongoing_alerts().json()
    all_alerts = pd.DataFrame(response)
    if all_alerts.empty:
        raise PreventUpdate
    live_alerts = all_alerts.loc[~all_alerts["is_acknowledged"]]

    return live_alerts.to_json()
```

Otherwise I've tested a bit, seems great to me ! 😄

app/homepage.py
Akilditu: The idea was to have the less number of elements within the repo. As this logo is also used and hosted for the website, let's not re-upload it.

Akilditu:
> Hi @pechouc nice PR, thank you very much. I've done some small comments.
> 
> There is an issue with the callback `update_alert_data` as you can see on this screen. I've investigated a bit and its linked to the fact that now the `response = api_client.get_ongoing_alerts().json()` returns an empty json (and the corresponding dataframe has no columns ...).
> 
> Capture d’écran 2021-04-01 à 09 29 21
> 
> To prevent this issue, you can change the callback to:
> 
> ```
> @app.callback(
>     Output("store_live_alerts_data", "data"),
>     Input('interval-component-homepage', 'n_intervals')
> )
> def update_alert_data(interval):
>     """
>     The following function is used to update the div containing live alert data from API.
>     """
>     # Fetching live alerts where is_acknowledged is False
>     response = api_client.get_ongoing_alerts().json()
>     all_alerts = pd.DataFrame(response)
>     if all_alerts.empty:
>         raise PreventUpdate
>     live_alerts = all_alerts.loc[~all_alerts["is_acknowledged"]]
> 
>     return live_alerts.to_json()
> ```
> 
> Otherwise I've tested a bit, seems great to me ! 😄

Agreed, debugged to the db reset 😅

app/main.py
pechouc: As seen during the call, @Akilditu will bear the heavy responsibility to erase everything 😢

app/main.py
pechouc: Done!

app/alerts.py
pechouc: Done, too!

requirements.txt
pechouc: I added a few lines in the README!

chloeskt (approved):
Great, thanks for the changes ! All good for me 👌

PR #44, opened 2021-03-30
Fix issue on alert screen

chloeskt:
Hi everyone ! 

Quick PR to fix a small bug that I found this morning, by pure chance. 

The issue: say a user opens the platform and then clicks on "écran de monitoring": he is now on the "/alert_screen" url. Then we clicks on "previous" on his browser: he's back on the homepage. Let's say now he wants to go back to "ecran de monitoring": there is an issue. Nothing is displayed on the screen.

I just fixed it: it was due to the fact that "last_displayed_event_id" was not reset when the user got back on the homepage. Now it is, which triggers correctly the code in the callback `update_alert_screen`.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -601,14 +601,18 @@ def callback_func_start_stop_interval(n_clicks):
     ],
     Input("interval-component-alert-screen", "n_intervals"),
     [
+        State("url", "pathname"),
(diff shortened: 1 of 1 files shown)


Akilditu (approved):
Thanks for the PR and correcting de bug !

PR #46, opened 2021-04-02
Fix latest issues with Dockerfile

chloeskt:
Hi all ! 

This PR aims at fixing the latest issues we got following #43, namely:

- add to `docker-compose.yml` the missing env variable LOGIN_CORRESPONDENCES_KEY
- add to the `Dockerfile` the instruction to copy the encrypted file (otherwise there is no file of logins to look up to !)

Additionnaly, in `main.py` I used `with open()` in order to avoid memory leak !!!! Indeed, in #43 we opened the file but never closed it ! It is easier to use with open since it automatically closes the file. + I added relative path (better than absolute path which could cause trouble).

Now if you rebuild the image, everything should work smoothly with Docker (at least on my hand it does).


Akilditu (approved):
Thanks for the PR, solved the login issues !

Not linked issue but it seems that exec time of the whole login process last 15 sec avg ...

PR #47, opened 2021-04-15
Login via api

pechouc:
# What is done

This pull request essentially modifies the `manage_login_modal` callback in `main.py`, so as to check the credentials entered by the user through the login route of the API instead of the encrypted login correspondences file. As a consequence, all references to the `simple-crypt` library in the Dockerfile, the requirements, etc can be eliminated. I also added `requests` to the requirements. 

# What remains to be done

I could not find a way to retrieve the `group_id` of the logged-in user. I tried the `users/me` or simply `users` routes in a notebook, which work just fine, but I could not find any `group_id` field in the output. I added placeholders in the code to fetch the group id properly via the API (right now, it is set arbitrarily to 1) and I also added a small `group_correspondences` dictionary, that will give for each group around what point to center the map, as well as the zoom level to use. 

Maybe I was looking in the wrong direction for the `group_id` 🤔


app/main.py
```diff
@@ -256,55 +262,48 @@ def manage_login_modal(n_clicks, username, password):
...
-        correspondences = pd.read_csv(correspondences)
+        login_route_url = 'http://pyronear-api.herokuapp.com/login/access-token'
 
-        if username not in correspondences['username'].values or password not in correspondences['password'].values:
-            # If either the username or the password is not found in the corresponding field of the login corresponden-
-            # ces csv file, the condition is verified
+        data = {
+            'username': username,
+            'password': password
+        }
```
Akilditu, line 274: It may sounds like a super dumb question, but when/where the fetch between the submitted username and pwd and this data object occurs ?

Akilditu (commented):
Thanks for this PR, should we wait for the `group_id` answer from @frgfm before merge ?

app/main.py
```diff
@@ -118,6 +115,16 @@
 alert_metadata = build_live_alerts_metadata()
 alert_id = alert_metadata["id"]
 
+# We define the coordinates of the point on which to center the map, as well as the appropriate zoom level for each
+# group of users (which will be used in the manage_login_modal callback below)
+group_correspondences = {
```
chloeskt, line 120: We will need to hardcode this ? This dictionnary is never stored anywhere else ?

app/homepage.py
```diff
@@ -358,7 +358,7 @@ def build_login_modal():
         id="login_modal",
 
         # Prevents the user from closing the modal by clicking next to it
-        backdrop='static',
+        # backdrop='static',
```
chloeskt, line 361: Should we keep this or remove it ?

app/main.py
pechouc: I am not sure to understand well the question but the username and password entered by the users are both a `State` of the callback, so they are passed as arguments to the `manage_login_modal` function. This allows to create the mini-dictionary `data` and to post it via `requests` on the next line. Is it what you wanted to know?

app/main.py
pechouc: If I understood well, what we will have for the tests in Ardèche on the API side is a way to fetch the `group_id` but there will not be fields like the center latitude and longitude and the zoom associated with it in the tables. So we would need to hardcode it, at least in a first version of the code (and as it will be very short, I guess it is simpler to hardcode it rather than storing it as a JSON file). But I might be wrong for the API and I am open to moving it to a static file if you prefer!

app/homepage.py
pechouc: Very nice catch, I commented this line for debugging but it should not stay as is indeed!

pechouc:
> Thanks for this PR, should we wait for the `group_id` answer from @frgfm before merge ?

Yes, I guess so! 

And otherwise, I have pushed a new commit to correct for the backdrop issue noticed by @chloeskt and to add some comments in the new pieces of code.

app/main.py
Akilditu: Oh well did not see the State ....

app/main.py
pechouc: Will do in the afternoon!

app/main.py
```diff
@@ -256,55 +262,48 @@ def manage_login_modal(n_clicks, username, password):
...
-            return True, form_feedback, '', [10, 10], 3
+            return True, form_feedback, [10, 10], 3
 
         else:
             # All checks are successful and we add the appropriate feedback
             # (although the login modal does not remain open long enough for it to be readable by the user)
             form_feedback.append(html.P("Vous êtes connecté, bienvenue sur la plateforme Pyronear !"))
 
+            # For now the group_id is not fetched, we equalize it artificially to 1
+            group_id = 1
```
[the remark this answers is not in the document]
frgfm: For now, in the same HTTP request: no
But I think on the API side, we're gonna add a "get my access" route : passing the token, it sends you back all non sensitive info (login, scope, group_id). Either way, consider that it will be done within the next days!

app/main.py
```diff
@@ -103,7 +102,7 @@
         dcc.Store(id="last_displayed_event_id", storage_type="memory"),
         dcc.Store(id="images_url_current_alert", storage_type="session", data={}),
         # Storage component which contains data relative to site devices
-        dcc.Store(id="site_devices_data_storage", storage_type="session")
+        dcc.Store(id="site_devices_data_storage", storage_type="session", data=get_site_devices_data(client=api_client))
```
Akilditu, line 105: So this `dcc.Store` allows one API call once the app is instantiated and then stores it on a session scope right ?

Akilditu (approved):
Thanks a lot for the PR ! Ready to merge :)

app/main.py
pechouc: Yes, that would be the plan but I am not an expert of these storage components, so I hope it works!

Issue #38, opened 2021-02-02
[API] Fetch alerts metadata from API


Akilditu:
2 things are still pending regarding `alerts` API calls : 

- [ ] Make actual calls to instantiate `alerts` metadata on some [remaining attributes ](https://github.com/pyronear/pyro-platform/blob/4a11c57dd82900f8eb6b0bc956125505f5eeed1b/app/utils.py#L192)
- [ ] Propose a relevant alternative to the 10 seconds scheduler or deal with backend ressources so that it never crashes

I would rather go for the web-hook option but I'm not sure that Dash handles this natively. Maybe we could use Flask endpoints that accept POST requests but then again I'm not sure that we could use it to trigger callbacks. 

Would be great to have your inputs on the 2nd point @frgfm @chloeskt @jeanpasquier75 @florianriche 👋

frgfm:
@florianriche mentioned webhook a few months back, but I think we need to identify the proper way to handle this.

First, the big choice: if we have 20 SDIS using the platform, should we deploy 1 server with multiple accesses, or 20 servers?

Now assuming we deploy only 1 server for the sake of efficiency: let's imagine we have 100 firefighters connected to the platform. We have two options here:
1. User independent: each user session is fetching its own data from the API according to the active creds/scope. Easier for data privacy, really bad for latency and server load.
2. Server-level: the platform server is receiving (via webhooks for instance) all the information from the API with appropriate scope, the user session will only receive the filtered information corresponding to its scope

This will largely impact future developments of the platform, so we need to decide first. My opinion is rather simple:
- I prefer having 1 server for the whole country (as long as latency is acceptable), and using server-level transfer. Mostly for consumption efficiency
- the minimum effort which preserves privacy without implementing much: 1 server per SDIS, and each user fetches its data. Easy, but not really responsible. Plus it will put additional constraints on the back-end team for scaling up the infra
What do you think?

frgfm:
Also, regarding your second point @Akilditu, I think we started an insightful discussion in #21

frgfm:
Not sure, but perhaps we can close #21 as both issues seem to be about the very same topic?

frgfm:
@florianriche would you mind reporting back here what you had found (resources / reference) about webhooks with flask? :pray:

frgfm:
Hello @pyronear/front-end :wave: 

[...]

PR #48, opened 2021-04-23
feat: New alert flow and websocket acting as a webhook to handle alerts date sent by API 

Akilditu:
### Motivation 
Hi Everyone ! This PR aims at initiating a new alert flow in regards of the UX team as well as implementing a websocket mechanism to handle API requests sent to the platform

### Features
**Websocket:**

As mentioned in #38 the idea was to move from frequent scheduled calls to the API to a webhook-like mechanism that could allow the API to send alert data each time an alert occurs.

Based on [WebSocket](https://pypi.org/project/dash-extensions/) the API is now able to ping the platform on a new /alert/message route(`message` being `str`) which then passes the value to an hidden div `msg` thanks to the `app.clientside_callback() `method.
This `msg`hidden div replaces the former `interval-component` as an input/trigger to the `update_live_alerts_data` callback , which then makes API calls retrieving both `live_alerts` and `dict_images_url_live_alerts` and storing them as outputs `dcc.Store` component.

`live_alerts` :  a json containing live alerts data where `is_acknowledged` is `False`
e.g `[{"id":16,"created_at":"2021-04-19T13:43:19.240131","media_id":25,"lat":44.8899993896,"lon":4.4099998474,"event_id":46,"is_acknowledged":false,"azimuth":214.6000061035,"device_id":1}]`

`dict_images_url_live_alerts` : a dict of list of live_alerts urls having event_id keys
e.g `{46: ['img_url_1, img_url_2'], 49: ['..']}`

The idea behind is that now pretty much every use case needing alert data can be fulfilled by those 2 `dcc.Store ` components ; as they are stored on a session level it's shared among the whole app.

**New alert flow:**

Based on the[ UI sketches](https://www.figma.com/file/OJxWB2CEcRm0cTlGTpH3Ne/Pyronear-User-flow-%F0%9F%8C%B3?node-id=26%3A16) designed by our UX team, the beginning of this new alert flow has been implemented. 
It implies changes and additions such as:

- deletion of the risk map entry
- navbar color changes
- navbar title changes while in alert mode
- 'full screen' map display
- alert banner reshaped into a new alerts button
- multiple alert markers
- a new user alerts selection area appearing after a click on new alerts button
- an alert list within the alerts selection area retrieving all live alerts

### Limits and remainings:

The websocket has some contraints such as: 

- The mandatory use of `run_server(app) ` which acts has a WSGI server wrapper upon Dash which does not allow the same debug smoothness. Anyone in dev mode will now have to restart the server instance after each Dash-related modification to see them appear
- To debug I recommend using either terminal for server-related logs and browser console for dash/front-related logs (keep in mind to restart instances to make the modifications appear

Still some work to do:

- Adding a button (or modifying the new alert one behavior) to be able to close user alerts selection area and come back to 'full map' mode
- Ending the full alert flow with alerts modals, redesign of alert markers till user feedbacks and alert acknowledgement (thanks for that @pechouc)
- Aligning the `alert_screen` related updates (mostly with the new dcc.Store components feeding the data needed). @chloeskt I did not make any heavy changes to the `alert_screen` but it seems that when 'Ecran de monitoring' is clicked an error occurs within browser console. Did not managed to solve that quickly, sorry for that ! 
- Ofc a bit of code harmonization, refactoring etc would be handy

Do not hesitate to reach out for any help, I assume that this PR is quite structuring so take time to review it !

Here is a snapshot of how it renders from login to the last implemented step of this new flow:
alert1
alert2


chloeskt:
Hello @Akilditu, thanks for this PR ! I will have a look at it this week-end hopefully and also try to see what's going on with the issue on "Ecran de monitoring" that you mentioned

app/alerts.py
```diff
@@ -330,15 +350,16 @@ def build_alerts_map():
     The following function mobilises functions defined hereabove or in the utils module to
     instantiate and return a dl.Map object, corresponding to the "Alerts and Infrastructure" view.
     """
-    map_object = dl.Map(center=[46.5, 2],     # Determines the point around which the map is initially centered
-                        zoom=6,               # Determines the initial level of zoom around the center point
+    map_object = dl.Map(center=user_department_lat_lon(),  # Determines the point around which the map is centered
```
pechouc, line 353: I am not sure to see the interest in having a specific function that defines the center and zoom of the map when instantiating it: as soon as the app is opened, callbacks are triggered that update these properties. Until the user is logged in, these properties are set to arbitrary values and then, they are modified to zoom the map around the right point. 

But I may be missing something here?

app/utils.py
```diff
@@ -35,6 +35,18 @@
 # ----------------------------------------------------------------------------------------------------------------------
 # CONTENT
 
+# ----------------------------------------------------------------------------------------------------------------------
+# User Info
+# The following block is used for the definition of logged user informations
+
+
+def user_department_lat_lon():
```
pechouc, line 43: I see where we use this function but as mentioned in the related comment, I am not sure whether it is useful.

app/utils.py
```diff
@@ -89,47 +101,31 @@ def choose_layer_style(n_clicks):
...
-
-    - the title of the information box and the name of the hovered when relevant;
-    - the title of the information box and a generic message when there is no department being hovered.
-    """
-    header_dept = [html.H4('Département sélectionné :')]
+# The following block is used to build the map filters in the bottom left corner
 
-    if feature:
-        return header_dept + [html.B(feature['properties']['nom'])]
+def build_map_filters(feature=None):
```
pechouc, line 106: Is this function still necessary if it simply returns the output of another? Should we add something here?

app/utils.py
```diff
@@ -89,47 +101,31 @@ def choose_layer_style(n_clicks):
...
     """
-    This function builds upon the build_info_box function defined above.
+    This function builds upon the build_map_filters function defined above.
 
     It takes as input the type of map considered (either 'alerts' or 'risks') and returns
-    an information box object located in the top-right corner of the map with a relevant id.
+    map filters of the map with a relevant id.
     """
     if map_type == 'alerts':
         object_id = 'alerts_info'
```
pechouc, line 119:
```suggestion
        object_id = 'alerts_filter_button'
```

Maybe we could adapt the ID of the object since it is not an information box anymore. Here is a proposition but it may not be the most appropriate one.

app/config.py
```diff
@@ -10,9 +10,8 @@
 load_dotenv()
 
 DEBUG: bool = os.environ.get('DEBUG', '') != 'False'
-API_URL: str = os.getenv('API_URL')
-API_LOGIN: str = os.getenv('API_LOGIN')
-API_PWD: str = os.getenv('API_PWD')
-LOGIN_CORRESPONDENCES_KEY: str = os.getenv('LOGIN_CORRESPONDENCES_KEY')
```
pechouc, line 16: Thanks for spotting and cleaning this!

app/homepage.py
```diff
@@ -389,22 +328,23 @@ def Homepage():
                 id='login_background',
                 children=[
                     # The background image is directly stored in the /assets folder
-                    html.Img(src='assets/background.png', width="100%")
+                    html.Img(src='assets/background.png', width="100%", height="100%")
```
pechouc, line 331: Regarding the issue that you pointed out (the fact that we can see the map even when not yet logged in, behind the modal), I will push a workaround in my next PR. The style attribute of the "hp_map" component is simply updated with `{'display': 'none'}` as long as the user is not logged in and with `{}` once login is successful, letting the map appear.

app/homepage.py
```diff
@@ -416,19 +356,19 @@ def Homepage():
                     html.Div(id='alert_btn_switch_view'),   # Associated with the alert banner in risks mode
 
                     # Simple placeholder - Source of truth for the map style being viewed
-                    html.Div(id='current_map_style', children='alerts'),
-
-                    # Hidden html.Div storing the URL address of the detection frame of the latest alert
-                    html.Div(id='img_url', style={'display': 'none'})],
-                    md=9),
+                    html.Div(id='current_map_style', children='alerts', style={'display': 'none'}),
+                    # Hidden div storing the webscocket message sent by the API
```
pechouc, line 360:
```suggestion
                    # Hidden div storing the websocket message sent by the API
```

Just a very small typo 😇

app/alerts.py
```diff
@@ -201,55 +194,54 @@ def build_vision_polygon(site_lat, site_lon, yaw, opening_angle, dist_km):
...
-            "iconUrl": '../assets/pyro_alert_icon.png',
-            "iconSize": [50, 50],        # Size of the icon
-            "iconAnchor": [25, 45],      # Point of the icon which will correspond to marker's and popup's location
-            "popupAnchor": [0, -20]      # Point from which the popup should open relative to the iconAnchor
-        }
-
-        # Building the list of alert markers to be displayed
-        alerts_markers = [dl.Marker(
-            id="alert_marker_{}".format(alert_id),   # Setting a unique id for each alert marker
+    # Changing the navabar color and title
```
pechouc, line 222:
```suggestion
    # Changing the navbar color and title
```

Again a really dumb comment about a small typo

app/alerts.py
```diff
@@ -201,55 +194,54 @@ def build_vision_polygon(site_lat, site_lon, yaw, opening_angle, dist_km):
...
+    for _, row in all_events.iterrows():
+        alert_id = str(row['event_id'])
+        alert_lat = row['lat']
+        alert_lon = row['lon']
+
+        alerts_markers.append(dl.Marker(
+            id={'type': 'alert_marker', 'index': alert_id},   # Setting a unique id for each alert marker
             position=(alert_lat, alert_lon),
             icon=icon,
             children=[dl.Popup(
```
pechouc, line 247: Thanks a lot for updating the IDs of the components of the marker popup! Although we probably kill the popup within the new alert workflow, right?

app/alerts.py
```diff
@@ -201,55 +194,54 @@ def build_vision_polygon(site_lat, site_lon, yaw, opening_angle, dist_km):
...
 
     In the base case, the function returns:
 
-    - the URL address of the image to be displayed on the left of the map;
-    - the alert button (banner above the map);
-    - the alert markers displayed on the map.
+    - the urls addresses of the image to be displayed on the left of the map
+    - the new_alert_button
+    - the alert markers displayed on the map
+    - new navbar_color as in alert mode
```
pechouc, line 215: I find this super super cool!

app/main.py
```diff
@@ -92,38 +97,43 @@
...
 
 # ----------------------------------------------------------------------------------------------------------------------
 # CALLBACKS
 
 # ----------------------------------------------------------------------------------------------------------------------
 # General callbacks
 
+
+# First API relay once an alert is sent to the platform, passing msg.data into msg hidden div
+app.clientside_callback("function(msg){if(msg == null) {return;} else {return msg.data;}}",
```
pechouc, line 133: Following our discussion, I think the reason why they write this callback as a [clientside callback](https://dash.plotly.com/clientside-callbacks) in the example we followed is in case the callback is triggered very often. If I understood correctly, those who wrote the extension are the same who rewrote the Dash source code to use WebSockets instead of HTTP requests to introduce interaction in the app. They were looking for performance gains and this should be allowed by the clientside callback. 

However, I feel like we do not really fall into the use cases described in the Dash documentation as the callback should not be triggered very often. I would thus advocate for transforming it in a plain Python callback. I mean not necessarily in this PR but I just wanted to raise this point and get your opinion on the matter!

app/main.py
```diff
@@ -275,8 +307,8 @@ def manage_login_modal(n_clicks, username, password):
         if not check:
             # This if statement is verified if credentials are invalid
 
-            # We add the appropriate feedback
-            form_feedback.append(html.P("Nom d'utilisateur ou mot de passe erroné."))
+            # We make the HTTP request to the login route of the API
+            response = requests.post(login_route_url, data=data).json()
```
pechouc, line 311: I think this was done by mistake, right? The POST request is done before the `if not check:` line.

Don't worry with this anyway, I will include this in my PR! 

(And do let me know if I am missing something of course)

app/main.py
```diff
@@ -310,6 +342,7 @@ def manage_login_modal(n_clicks, username, password):
 
 @app.callback(
     Output('login_background', 'children'),
+    Output('main_navbar', 'style'),
```
pechouc, line 345: I hadn't noticed that the navbar was hidden until login, nice catch!

app/main.py
```diff
@@ -323,29 +356,12 @@ def clean_login_background(is_modal_opened):
...
-
-    This callback detects what department is being hovered by the user's cursor.
-
-    It returns the corresponding name in the info object in the upper right corner of the map.
-    """
-    if hovered_department is not None:
-        return build_info_box(hovered_department)
-    else:
-        return build_info_box()
-
```
pechouc, line 348: 😢

pechouc (commented):
Hello! 

Thanks a lot for this massive PR and for dealing with all the bugs until everything worked smoothly 🙏 

Overall, I find it very smooth and my comments are mostly anecdotal or raising a few points for future discussions. I simply feel like the `utils.py` file could be simplified: in particular, I am not sure about the utility of the `user_department_lat_lon` function and we could easily reduce the process to build the filters on the map down to 1 or 2 functions.

I hope that everything will be clear and again, many thanks 😊

app/alerts.py
Akilditu: Oh yeah you're right, did that before the login PR

app/utils.py
Akilditu: Indeed, done !

app/homepage.py
Akilditu: Alright, great !

app/homepage.py
Akilditu: Thanks!

app/alerts.py
Akilditu: Indeed, did that to handle and test multiple alerts markers

app/utils.py
Akilditu: Done

app/main.py
Akilditu: Why not, I honestly don't have any strong opinion on this but I agree that for code unity it might be better to use 100% Python callbacks.

Ok for changing it in a next PR !

app/main.py
Akilditu: Yeah this might be a conflict miss resolution while merging

app/main.py
Akilditu: I know .... This could be reverted if necessary in the future 😅

Akilditu (commented):
@pechouc thanks for your feedback, just implemented last changes !

pechouc:
Perfect, thanks a lot, we are good to go then!

pechouc (approved):

PR #50, opened 2021-05-04
Import fix to deploy

Akilditu:
This PR aims at fixing a tiny import fail while deploying on OVH


pechouc (approved):
Looks good to me, thanks for spotting and correcting this!

PR #53, opened 2021-05-10
Alert workflow and acknowledgement

pechouc:
Hello! 

This PR focuses mainly on the alert workflow, building upon what has already been merged with the Websocket etc. 

## Included in this PR

It introduces: 

- the vision angles of devices having raised an alert;
- the zoom on the relevant alert when the user clicks on any of the alert boxes on the left-hand side of the map;
- the alert overview that appears on the left-hand side of the map when clicking on one of the alert boxes;
- the alert acknowledgement mechanism that can be triggered from the alert overview;
- the alert modal accessible by clicking on the small alert frame on the left-hand side of the map;
- the image gallery, to view the different frames associated with the same alert/event. 

Additionally, this PR modifies the filtering of live alerts. These are not identified solely based on acknowledgement or not but as the alerts that either correspond to an unacknowledged event or that have been created less than 12 hours ago. 

Third, this PR corrects a few unwanted behaviours in the login, especially to hide the map as long as the user has not provided valid credentials. 

Finally, it introduces a few very small changes here and there, mostly in the comments. 

## Not included in this PR

Several elements are not covered by this PR: 

- the button to hide an irrelevant alert throughout the session; 
- a mechanism that would avoid opening the login modal each time the page is refreshed during a given browser session; 
- re-establishing the link with the big alert screen with the top-right button;
- the alert feedback mechanism (where to store the answers of the user...);
- getting closer to the Figma design.

This is a top-of-the-head list and there are probably some elements that are still missing!

## Screenshots and testing with actual data

I lamentablement failed to create a new alert via the API client or directly via a POST. 

As a consequence, because the DB has been emptied, I do not have interesting screenshots to share and I could not actually test the acknowledgement mechanism ☹️

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -245,7 +245,7 @@ def update_live_alerts_data(alert):
 
     # We first want to build the boolean indexing mask that corresponds to the time-related condition
     # We convert values in the "created_at" column to datetime format
-    all_alerts['created_at'] = pd.to_datetime(all_alerts['created_at'])
+    all_alerts['created_at'] = pd.to_datetime(all_alerts['created_at'], utc=True)
 
     # For each of these creation dates, we check whether they were registered less than 12 hours ago
     # This provides us with the first boolean indexing mask
@@ -255,7 +255,7 @@ def update_live_alerts_data(alert):
     # We now want to build the boolean indexing mask that indicates whether or not the event is unacknowledged
     # We start by making an API call to fetch all events
     url = cfg.API_URL + '/events/'
-    all_events = requests.get(url, headers=api_client.headers)
+    all_events = requests.get(url, headers=api_client.headers).json()
 
     # Then, we construct a dictionary whose keys are the event IDs (as integers) and values are the corresponding
     # "is_acknowledged" field in the events table (boolean)
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,5 @@ pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirect
 python-dotenv>=0.15.0
 geopy>=2.1.0
 requests>=2.25.0
+pytz>=2020.1
+


app/homepage.py
```diff
@@ -339,7 +341,11 @@ def Homepage():
             [
                 # Left column containing the user selection area
                 dbc.Col([
+                    # At first, the map is hidden (until the user logs in with valid credentials)
                     html.Div(build_user_selection_area(), id='selection_area', style={'display': 'none'}),
+
+                    html.Div(id="alert_overview_area"),
+
```
Akilditu, line 348: Thanks for adding this !

app/homepage.py
```diff
@@ -349,15 +355,20 @@ def Homepage():
...
                     html.Div(id='map_style_btn_switch_view'),   # Associated with the main map style button
                     html.Div(id='alert_btn_switch_view'),   # Associated with the alert banner in risks mode
 
                     # Simple placeholder - Source of truth for the map style being viewed
                     html.Div(id='current_map_style', children='alerts', style={'display': 'none'}),
-                    # Hidden div storing the websocket message sent by the API
+
+                    # Two placeholders updated by callbacks in main.py to change center and zoom attributes of the map
+                    html.Div(id='login_zoom_and_center', style={'display': 'none'}),
+                    html.Div(id='alert_zoom_and_center', style={'display': 'none'}),
```
Akilditu, line 369: Nice ways to update center and zoom from several triggers

app/alerts.py
```diff
@@ -304,25 +467,57 @@ def build_user_alerts_selection_area(n_clicks, live_alerts):
...
-            alert_azimtuth = round(row['azimuth'], 1)
+            alert_azimuth = round(row['azimuth'], 1)
             alert_date = datetime.fromisoformat(str(row['created_at'])).date()
             alert_time = datetime.fromisoformat(str(row['created_at'])).time()
-            alert_list.append(html.Div([
+
+            alert_selection_button = html.Div([
                 dcc.Markdown('---'),
                 dbc.Button(children=[
-                           html.Span('Azimuth : {}°'.format(alert_azimtuth), style={'font-weight': 'bold'}),
```
Akilditu, line 318: thanks for the catch

app/alerts.py
```diff
@@ -144,7 +144,7 @@ def build_sites_markers(dpt_code=None):
     return dl.MarkerClusterGroup(children=markers, id='sites_markers')
 
 
-def build_vision_polygon(site_lat, site_lon, yaw, opening_angle, dist_km):
+def build_vision_polygon(event_id, site_lat, site_lon, yaw, opening_angle, dist_km):
```
Akilditu, line 147: Great improvement !

app/alerts.py
```diff
@@ -281,19 +269,194 @@ def build_alerts_elements(images_url_live_alerts, live_alerts, map_style):
...
+                                                id={
+                                                    'type': 'alert_type_radio_button',
+                                                    'index': str(event_id)
+                                                },
+                                                options=[
+                                                    {'label': 'Feu de forêt', 'value': 'wildfire'},
+                                                    {'label': 'Incendie domestique', 'value': 'domestic_fire'},
+                                                    {'label': 'Cheminée', 'value': 'chimney'},
+                                                    {'label': 'Nuage', 'value': 'cloud'},
+                                                    {'label': 'Autre', 'value': 'other'}
```
Akilditu, line 404: To be challenged but good first draft !

app/main.py
```diff
@@ -201,30 +224,70 @@ def update_live_alerts_data(alert):
...
         raise PreventUpdate
-    live_alerts = all_alerts.loc[~all_alerts["is_acknowledged"]]
+
+    # We store all alerts in a DataFrame and we want to select "live alerts",
+    # Ie. alerts that either correspond to a not-yet-acknowledged event or have been created less than 12 hours ago
+    all_alerts = pd.DataFrame(response)
+
+    # We first want to build the boolean indexing mask that corresponds to the time-related condition
+    # We convert values in the "created_at" column to datetime format
+    all_alerts['created_at'] = pd.to_datetime(all_alerts['created_at'], utc=True)
```
Akilditu, line 248: Nice catch !

app/main.py
```diff
@@ -201,30 +224,70 @@ def update_live_alerts_data(alert):
...
+    # "is_acknowledged" field in the events table (boolean)
+    is_event_acknowledged = {}
+    for event in all_events:
+        is_event_acknowledged[event['id']] = event['is_acknowledged']
+
+    # We map this dictionary upon the column and revert the booleans with ~ as we want unacknowledged events
+    mask_acknowledgement = ~all_alerts['event_id'].map(is_event_acknowledged)
+
+    # We link the two masks with an OR condition
+    mask = np.logical_or(mask_time, mask_acknowledgement)
```
Akilditu, line 270: Shouldn't we want an AND condition here ? Meaning that live_alerts are those referring an unacknowledged event lasting for more that 12 hours ?

Akilditu (changes requested):
Great improvements overall regarding alert flow and alert acknowledgment ! 

I was thinking that before merging you could also add the prod API creds in .env in a dedicated commit !

app/main.py
pechouc: With an AND, an alert/event would disappear from the platform as soon as it is acknowledged (or to be more precise, at the next update of the alert data), which we wanted to avoid by adding a time condition. So, I do believe that this should be an OR, with three possibilities for an alert to be displayed: 

- Either it is still "fresh" as it was created less than 12 hours ago;
- Or it has not yet been acknowledged by firemen (how can we be sure that they taken it into account otherwise?);
- Or both. 

Let me know if this remains unclear 😊

pechouc:
Thanks a lot for all your comments @Akilditu! However, I am not sure to understand what you mean with the .env file, do we really want to push the API credentials on GitHub? 🤨

Akilditu (approved):
Thanks a lot !

PR #54, opened 2021-05-24
Alert erasing button

pechouc:
Hello everyone! 

# Brief description of the PR

This PR aims at introducing two new features in the overview area of each alert/event:

- a button to close the overview and come back to the selection of the alerts;
- a button to erase an alert from the platform for the rest of the browser session. 

These simple interactions required a bit of work, essentially in `main.py`, as several outputs are now determined by a multitude of inputs (not only the center and zoom attributes of the map anymore). It also required to modify the `build_individual_alert_components` in `alerts.py`, some of its outputs being transferred to a new function (`display_alert_selection_area`). 

I have tried to comment the code as much as possible but let me know if you have any question of course!

# Screenshots

- Two alerts are initially displayed on the platform

Capture d’écran 2021-05-24 à 08 53 46

- As usual, we click on the alert button in the top-right corner of the map and then, select one of the alerts which lets the overview appear on the left-hand side of the map

Capture d’écran 2021-05-24 à 08 54 16

- By clicking on "Fermer l'aperçu de l'alerte", one can close the alert overview 

Capture d’écran 2021-05-24 à 08 55 11

- After reopening it (cf. second screenshot) and clicking on "Ne plus voir cette alerte", only 1 alert is displayed by the platform now

Capture d’écran 2021-05-24 à 08 56 05

- We can open the remaining alert overview

Capture d’écran 2021-05-24 à 08 56 57

- And we can do the same for this alert, no alert being displayed on the platform (back to its initial, no-alert state)

Capture d’écran 2021-05-24 à 08 57 38

# Question

I don't know if you have the same issue but as you can see on the screenshots, the map doesn't load properly when I build the app locally with Docker. Using `python3 app/main.py" works (slightly) better and the deployed platform seems to display the map correctly, so it should not be too big an issue. But still, it does not help with the debugging...

### DIFF
--- a/app/alerts.py
+++ b/app/alerts.py
@@ -246,6 +246,18 @@ def build_alerts_elements(images_url_live_alerts, live_alerts, map_style):
     # Building the list of alert markers to be displayed
     alerts_markers = []
     all_alerts = pd.read_json(live_alerts)
+
+    if all_alerts.empty:
+
+        hidden_header_alert_button = html.Div(
+            html.Button(
+                id=f'alert_button_{map_style}'
+            ),
+            style={'display': 'none'}
+        )
+
+        return [hidden_header_alert_button], [], '#054546', 'Surveillez les départs de feux', []
+
     all_events = all_alerts.drop_duplicates(['id', 'event_id']).groupby('event_id').head(1)  # Get unique events
     for _, row in all_events.iterrows():
         alert_id = str(row['event_id'])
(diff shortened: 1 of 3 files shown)


Akilditu (approved):
Hi @pechouc, great PR everything's looks good to me ! 

Regarding your question I do have the same issue and don't know why. Let's try to fix this later on.

PR #57, opened 2021-06-19
Reviewing the early alert workflow

pechouc:
This PR aims at reviewing the early alert workflow, i.e. mainly the callback that is triggered by the Dash `Interval` component, loads alerts data via the API client and updates the dedicated storage components. In particular, this updated version of the process limits as much as possible the number of calls to the `get_media_url` method of the client, which can be a bit slow. Besides, the length of the interval is variabilized: since the first load can be quite long, the initial time span is set at 15 seconds but following executions being much faster, we move the interval length to 5 seconds (which minimises the delay to display a new event or a new frame on the platform).


app/main.py
```diff
@@ -371,32 +399,179 @@ def update_live_alerts_data(n_intervals, ongoing_live_alerts, ongoing_frame_urls
...
+
+            # We drop the azimuth associated with the alert as we will focus on the yaw of the device
+            live_alerts = live_alerts.drop(['azimuth'], axis=1)
+
+            # We rename columns to avoid any ambibguity (id_y is the id of the device)
+            live_alerts.rename(columns={'id_x': 'id', 'id_y': 'd_id'}, inplace=True)
+
+            # We store the IDs of newly loaded alerts in a dedicated list
+            # This will serve as the source of truth to know what frame URLs have already been fetched or not
+            new_loaded_frames = list(live_alerts['id'].unique())
```
Akilditu, line 467: Handy !

app/main.py
```diff
@@ -371,32 +399,179 @@ def update_live_alerts_data(n_intervals, ongoing_live_alerts, ongoing_frame_urls
...
-            img_url = api_client.get_media_url(row["media_id"]).json()["url"]
-        except Exception:
-            pass
+        # We load the data contained by the store_live_alerts_data dcc.Store component
+        temp = json.loads(ongoing_live_alerts)
+
+        # Has some alert data already been loaded so far?
+        if isinstance(temp, dict) and 'status' in temp.keys() and temp['status'] == 'never_loaded_alerts_data':
+            # If the dictionary loaded from the store_live_alerts_data dcc.Store component corresponds to the initial
+            # data attribute of the component, it means that alert data have never been loaded so far
```
Akilditu, line 422: Nice way to check that the initial store_live_alerts_data has been fed with alert data in the first place !

app/main.py
```diff
@@ -371,32 +399,179 @@ def update_live_alerts_data(n_intervals, ongoing_live_alerts, ongoing_frame_urls
...
+                    ]
+
+                # If the condition is not verified, we have no new "alert" / event to display on the platform but only
+                # new detection frames for an existing alert; this means that we do not have to update all components
+                else:
+
+                    # We would like to only update the list of alert frames being displayed and not all the components
+                    # To keep track of the frame URLs that have been loaded, we also update the list of loaded alert IDs
+                    return [
+                        dash.no_update,
```
Akilditu, line 570: Nice dash.no_update for a specific output, still impressed of endless dash framework deepness 😅

app/main.py
```diff
@@ -306,31 +329,36 @@ def update_live_alerts_data_erase_buttons(n_clicks, alerts_data, alerts_frames):
...
-    - a dict where keys are event id and value lists of all urls related to the same event.
-
-    These URLs come from the API calls, triggered by the intervals every 5 seconds.
+    This is the key callback of the platform. Triggered by the interval component, it queries the database via the API
+    client so as to identify live alerts, load the associated data and trigger their display on the platform. This doc-
+    string should be completed but more details can be found in the comments below.
     """
 
+    print('Number of intervals:', n_intervals)
+    print('Time:', time.time())
```
Akilditu, line 360: Do we actually need that in production ?

app/main.py
```diff
@@ -371,32 +399,179 @@ def update_live_alerts_data(n_intervals, ongoing_live_alerts, ongoing_frame_urls
...
+            live_alerts = live_alerts.to_json(orient='records')
+
+            # Reminder: in this case, this is the first time that we load alert data
+
+            # So we update all outputs:
+            # - the storage component that contains alert data in JSON format;
+            # - the storage component that contains the dictionary with detection frame URLs;
+            # - the storage component that serves as source of truth for the list of already loaded alerts
+
+            return live_alerts, dict_images_url_live_alerts, {'loaded_frames': new_loaded_frames}, 5 * 1000
```
Akilditu, line 479: Switching from 15 seconds to 5 seconds seems fair, any idea on how you decided to setup 15 seconds for the first -lazy- callback exec (seems safe as well though) ?

Akilditu (commented):
Thanks a lot for this PR ! 

Nice refactoring + a lot of performance and loading time optimization. Left mostly dumb comments and will experiment it a bit more before approval !

app/main.py
pechouc: No, you're right, I forgot to remove it!

pechouc: Done 😊

app/main.py
pechouc: Typically, with the live alerts that are currently in the DB, the first callback takes a bit more than 10 seconds if I am not mistaken. So I thought it was important to take some safety margin although I doubt we will have as many live alerts as right now once deployed on the field (difficult to anticipate actually because firemen will probably use the acknowledgement button but at the same time, we can have many frames following each other for the same event...). In any case, I feel like leaving some comfortable loading time for the first callback to run should not be too much of an issue as long as the following callbacks are shorter and executed more often!

app/main.py
```diff
@@ -195,10 +213,11 @@ def change_layer_style(n_clicks=None):
 
 @app.callback(
     Output('store_live_alerts_data', 'data'),
-    [Input('update_live_alerts_data_workflow', 'children'),
-     Input('update_live_alerts_data_erase_buttons', 'children')]
+    [Input('update_live_alerts_data_workflow', 'data'),
+     Input('update_live_alerts_data_erase_buttons', 'data')],
+    State('update_live_alerts_data_workflow', 'data')
```
Akilditu, line 218: There must be smth but I don't see the need of having `update_live_alerts_data_workflow` both as Input and State here ?

Akilditu (commented):
What would you think of disable the debug mode for PROD. I guess that a quick but maybe dirty solution would be to force `app.run_server(host=args.host, port=args.port, **debug=False**`.

app/main.py
pechouc: You're right, I don't know why it was there in the first place!

Akilditu (approved):
Thanks a lot for the PR and updates !

frgfm:
@Akilditu btw, don't forget to delete the branch once the PR is merged :sweat_smile:

PR #58, opened 2021-06-24
Small debugging

pechouc:
The aim of this PR is to: 

- Correct for the error message that says "A nonexistent object was used in an `Output` of a Dash callback. The id of this object is `update_live_alerts_data_workflow` and the property is `data`." and currently appears when opening the platform locally;

- Correct for bugs that appear when there is only one alert / event being displayed and one clicks on the "Ne plus voir cette alerte" button. These are related to the way the `build_alerts_components` and `build_individual_alerts_components` functions were written in `alerts.py`.

**Warning:** It might be that the first change (moving placeholders from `homepage.py` to `main.py`) creates another error message, which I have seen a few times but cannot recreate anymore. It was related to the `select_alert_frame_to_display` callback and basically said that the `individual_alert_frame_storage` component associated with the alert / event being displayed had not been created before being used as a `State`. This does not prevent the alert from being displayed on the platform but no image appears in the alert modal when opening it for the first time. 

@Akilditu, could you please counter-check this debugging proposal and see if you observe the error message locally (the one described in the warning)? 🙏

### DIFF
--- a/app/alerts.py
+++ b/app/alerts.py
@@ -247,8 +247,6 @@ def build_alerts_elements(images_url_live_alerts, live_alerts, map_style, blocke
     alerts_markers = []
     all_alerts = pd.read_json(live_alerts)
 
-    all_alerts = all_alerts[~all_alerts['event_id'].isin(blocked_event_ids['event_ids'])].copy()
-
     if all_alerts.empty:
         # When there is no live alert to display, we return a alert header button that will remain hidden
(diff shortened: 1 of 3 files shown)


Akilditu (approved):
Thanks for the PR ! 

Juste tested and no problems seem to occur, let's merge if there is no conflicts due to previous PR.

PR #60, opened 2021-06-28
Various edits for demo

pechouc:
Hello, 

This PR aims at: 

- enabling a simplified but a priori functional version of the big alert screen;
- (temporarily?) disabling the display of historic fire markers;
- extending the delay for loading the initial alerts data when the app is opened;
- displaying the site name that corresponds to each alert everywhere on the platform. 

Thanks in advance for the review!


Akilditu (approved):
Thanks for the PR !

PR #63, opened 2021-07-10
feat: Preventing alerts to be displayed

Akilditu:
This PR disables alert display for demo purposes only. This is done my emptying the response from API alert fetches and will be removed later on in production.


pechouc:
Hello, thanks a lot for this PR and for the work on refreshing the API client! 

Shouldn't we add a similar line to the callback that builds the alert screen so that we don't display alerts on it either?

Akilditu:
> Shouldn't we add a similar line to the callback that builds the alert screen so that we don't display alerts on it either?

Indeed, done !

pechouc (approved):
Perfect, thanks a lot for the edit!

PR #65, opened 2021-08-02
feat: Filtering night alerts out of plateform

Akilditu:
This PR aims at preventing alerts that may occur between sunset and sunrise to be displayed on the platform for now as our model is not tailored for night detection yet. 

Those are the main steps : 

- One the platform is instantiated a call is made to `api.sunrise-sunset ` to fetch and store sunrise and sunset data within a `dcc.Store` component
- The function `is_hour_between` is defined to return True if a specific time happens between sunrise and sunset
- The API data is passed as a State of the main callback `update_live_alerts_data`, converted from UTC to GMT + 2 (with a 'safe margin' of 30 min for both sunrise and sunset times)
- Finally we iterate over `live_alerts` and if `created_at` alert time is located between sunset and sunrise then we drop the row 

Remaining and enhancements : 

- Today a central `lat lon` has been taken in the middle of fire towers in Ardeche, fore more accuracy we could give the proper `lat lon` of every tower
- The call to API is done every time the app is instanciated, we may want a dedicated callback to refresh this ervery 24h


app/main.py
```diff
@@ -158,13 +165,29 @@
     'CACHE_DEFAULT_TIMEOUT': 60
 })
 
+# ----------------------------------------------------------------------------------------------------------------------
```
pechouc, line 168: I would rather define this function in `utils.py` and simply import it in `main.py`, what do you think?

app/main.py
```diff
@@ -158,13 +165,29 @@
...
+
+
+def is_hour_between(sunrise, sunset, alert_time):
+
+    alert_time = datetime.fromisoformat(str(alert_time)) + timedelta(hours=2)
+    alert_time = alert_time.time()
+
+    is_between = False
+    is_between |= sunrise <= alert_time <= sunset
+    is_between |= sunset <= sunrise and (sunrise <= alert_time or alert_time <= sunset)
```
pechouc, line 180: Could you please explain in what case `sunset <= sunrise` evaluates to `True`?

app/main.py
```diff
@@ -289,6 +314,23 @@ def update_live_alerts_data(
     # And we deduce the subset of alerts that we can deem to be "live"
     live_alerts = all_alerts[mask_acknowledgement].copy()
 
+    # We then fetch sunrise and sunset times and add a safety margin of 30 min (converting from UTC) to cover night time
+    sunrise = night_time_data['results']['sunrise'][:-6]
+    sunrise = datetime.fromisoformat(str(sunrise)) + timedelta(hours=2.5)
```
pechouc, line 319: If the hours returned by `api.sunrise-sunset` are in UTC too, wouldn't it be simpler not to make any conversion to GMT (neither in the `is_hour_between` function nor here)?

app/main.py
```diff
@@ -289,6 +314,23 @@ def update_live_alerts_data(
...
+    sunrise = sunrise.time()
+
+    sunset = night_time_data['results']['sunset'][:-6]
+    sunset = datetime.fromisoformat(str(sunset)) + timedelta(hours=1.5)
+    sunset = sunset.time()
+
+    # Are there some live alerts during night time ? If yes let's filter them out
```
pechouc, line 326: I would rather tend to use a `.map` and boolean indexing here. I am not a specialist of performance matters but it seems to me than it is generally faster than writing `for` loops. This could give something like: 

```
mask = live_alerts['created_at'].map(lambda x: is_hour_between(sunrise, sunset, x))
live_alerts = live_alerts[mask].copy()
```

Let me know what you think about this 🙂

app/main.py
```diff
@@ -131,6 +132,12 @@
         # [NOT SUCCESSFUL YET]
         dcc.Store(id='login_storage', storage_type='session', data={'login': 'no'}),
 
+        # Session storage component which fetches sunrise and sunset times to filter night alerts
+        dcc.Store(id='night_time',
+                  storage_type='session',
+                  data=requests.get('https://api.sunrise-sunset.org/json?lat=44.62112179704533n'\
+                                    '&lng=4.273138903528911&formatted=0&date=today').json()),
+
```
pechouc, line 140:
```
# Updates the sunrise and sunset data every 12 hours
dcc.Interval(id='refresh_night_time', interval=12 * 3600 * 1000),
```

app/main.py
```diff
@@ -158,13 +165,29 @@
...
+    is_between |= sunrise <= alert_time <= sunset
+    is_between |= sunset <= sunrise and (sunrise <= alert_time or alert_time <= sunset)
+
+    return is_between
 
 # ----------------------------------------------------------------------------------------------------------------------
 # CALLBACKS
 
 # ----------------------------------------------------------------------------------------------------------------------
 # General callbacks
```
pechouc, line 189:
```
@app.callback(Output('night_time', 'data'), Input('refresh_night_time', 'n_intervals')
def refresh_night_time(n_intervals):
   return requests.get('https://api.sunrise-sunset.org/json?lat=44.62112179704533n'\
                                     '&lng=4.273138903528911&formatted=0&date=today').json()
```

pechouc (commented):
Thanks a lot for this PR, I have left a few comments!

For the callback used to refresh sunrise and sunset data, it is an excellent idea and I think this should be part of the PR. I have left suggestions to spare you some time in the comments 😊

As for `lat` and `lon`, I am not sure to understand the issue. Are we not displaying the alerts at the right spot (the coordinates of the tower from which the alert has been sent)?

app/main.py
Akilditu: Done

app/main.py
Akilditu: You're right but as we need to use timedelta to add the 30 min margin this is not an issue to leave it like this i guess

app/main.py
Akilditu: Thanks for this indeed it seems to be more efficient !

app/main.py
Akilditu: Indeed, done !

app/main.py
Akilditu: Thanks, done !

app/main.py
Akilditu: In a rare case of working with UTC let's say in Hawai or Japan it may occur that sunset <= sunrise turns to be True. Agree that won't never be the case for us but leaving it there

app/main.py
pechouc: Many thanks!

app/main.py
pechouc: Well noted, thanks for the indication!

pechouc:
All good on my end, thanks a lot for your answers and edits! Do you think you could quickly correct for the unused import raised by Codacy?

pechouc:
Ahah perfect, I will validate the PR as soon as checks have passed 😊

pechouc (approved):

PR #68, opened 2021-08-02
Update requirements.txt

pechouc:
Trying to modify the `pyroclient` requirement so that the latest version of the package is installed on the Heroku machine.


Akilditu (approved):
Thanks for the PR !

Issue #71, opened 2022-01-14
[map] The OpenStreetMap background is either offset or has a problem


frgfm:
Apparently, there is an issue with the alert view: the background map is not displayed correctly.

Here is how it renders:
![image](https://user-images.githubusercontent.com/26927750/149569330-d13871ed-1fea-4281-a146-c7dcc3ce7a72.png)

I've heard suggestions it's because of a recent release of one of the dependencies (probably the ones related to OpenStreetMap). Could someone have a look? :pray:

PR #72, opened 2022-01-20
update dependance of the projet ( dash also ) , find the bug and find a work around 

TimotheDavid:
Hey I have fix the dep ( why it's doesn't work :) now it's dash > 2 
I ahve find a work around for the bug about the screen ( resize the screen ( yess it's work !) 

thnaks for accpet my pr

### DIFF
--- a/app/alert_screen.py
+++ b/app/alert_screen.py
@@ -82,7 +82,7 @@ def build_alert_detected_screen(img_url, last_alert, site_devices_data):
     fire_icon = html.Img(
         id="fire_icon",
         src="/assets/pyro_fire_logo.png",
-        className="blink-image",
+        className="blinkImage",
         style={"height": "100%"},
     )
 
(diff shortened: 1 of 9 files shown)


Akilditu (commented):
Hi @TimotheDavid, sorry for the late concern, few questions here :

- The motivation of this PR is to provide a fix for the force resize of the browser tab that was needed to display the map right ? 
- The typo changes that you did are requested by a recent dash version update ?

Thanks for the PR !

frgfm:
> Hi @TimotheDavid, sorry for the late concern, few questions here :
> 
> * The motivation of this PR is to provide a fix for the force resize of the browser tab that was needed to display the map as explained [here](https://github.com/pyronear/pyro-platform/issues/71) right?
> * The typo changes that you did are requested by a recent dash version update ?
> 
> Thanks for the PR !

If specifying a max version of dash does the trick, perhaps that would be easier :+1:

frgfm:
> If specifying a max version of dash does the trick, perhaps that would be easier +1

What do you think @TimotheDavid? :)

Akilditu:
> > If specifying a max version of dash does the trick, perhaps that would be easier +1
> 
> What do you think @TimotheDavid? :)

Hi definitely think that it's the best solution !

frgfm:
Hello everyone :wave: 

Reading the conversation again, I think we have to use the version specifiers to solve this problem. I opened a dedicated issue for this #76, so I'll close this PR.

Feel free to reopen if you think it's relevant :)

PR #73, opened 2022-05-03
Fix: Filtering fire alerts to reduce platform loading time

Akilditu:
This PR aims at : 

- Trying to lighten ressources and loading time displaying only the last 5 alerts
- Align dashboard devices `last_ping` timezone to FR

Enhancements & thoughts : 

- Still long to load while local testing this filters might be more relevant on API side @frgfm ? 
- Parsing code / doing unit testing on API fetches could be an option to determine more precisely what's actual costing time ? @pechouc ? 

Thanks feedback and review


frgfm:
Hey there 👋 

So in the long-term, we will need to reduce the workload the platform imposes on the API. I will add a result limit (can only GET the last N events) on most API routes.

REgarding the platform, the CI isn't testing the modifications you proposed here so here is a suggestion:
- we make independent functions to perform those operations
- we add unittests for those
- we use the functions in the actual code

That could be done in a PR later but we should do it at some point 😅

pechouc (commented):
Hello, sorry for the late answer!

The change looks OK to me and we can merge if you want to move forward with this solution. 

Regarding the loading times and testing, I fully agree that we need to work on these. We could dedicate more time after the Data for Good demo-day I guess, but I have prepared a small (and very ugly) piece of code that allows to decompose each step and measure running times. I have run it on my computer and saved loading times. 

In essence, the message is the following:

- the API call through the `get_ongoing_alerts` method takes around 3 seconds in a rather consistent way;

- the call to the `events` route of the API adds half a second approximately;

- DataFrame manipulations are close to insignificant in terms of computation time;

- when creating the objects for the first time, the big `for` loop that starts around line 400 of the file and notably fetches alert frame URLs explains most of the running time. It seems to take 5 seconds when restricting to the last 5 events (hence the fact that the restriction is likely to eliminate running time issues) and more than 80 seconds when I removed the filtering! This part is only executed when no alert has been loaded before, so this happens only once.

I have written the code in another branch and saved the loading times in JSON files. There is also a small notebook to read these. Let me know if you want to check these and if so, what would be the easiest way to share them.

frgfm:
So FYI everyone, we have already merged on the backend API side, the modification to only be able to fetch the last 50 entries instead of loading the whole table. That should speed things considerably

For the rest, including the "for" loops, I'll have to take a closer loop if we can optimize that a bit. But we for sure need to move forward with this PR

Akilditu:
> So FYI everyone, we have already merged on the backend API side, the modification to only be able to fetch the last 50 entries instead of loading the whole table. That should speed things considerably
> 
> For the rest, including the "for" loops, I'll have to take a closer loop if we can optimize that a bit. But we for sure need to move forward with this PR

Hi @frgfm thanks a lot for this, could you confirm the impacted route ?

app/main.py
```diff
@@ -334,6 +334,11 @@ def update_live_alerts_data(
     else:
         # If yes, there is a bit of work to do!
 
+        # Let's only display the last 5 fire events!
+
+        last_5_events = live_alerts['event_id'].drop_duplicates().sort_values().tail(5)
+        live_alerts = live_alerts[live_alerts['event_id'].isin(last_5_events)]
```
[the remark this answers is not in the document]
Akilditu: Thanks @frgfm, will test and commit this :)

frgfm:
> Hi @frgfm thanks a lot for this, could you confirm the impacted route ?

All the fetch routes (not update, delete, or get) benefit from this :+1: considering the size of tables, fetching media, alerts or events should be much faster now

frgfm (dismissed):
Looks good but you need to merge master into this branch to resolve the conflicts :)

app/main.py
```diff
@@ -293,24 +294,15 @@ def update_live_alerts_data(
...
-        response = api_client.get_past_events()
-    all_events = response.json()
-
-    # Then, we construct a dictionary whose keys are the event IDs (as integers) and values are the corresponding
-    # "is_acknowledged" field in the events table (boolean)
-    is_event_acknowledged = {}
-    for event in all_events:
-        is_event_acknowledged[event["id"]] = event["is_acknowledged"]
+        response = api_client.get_unacknowledged_events()
+    all_events = pd.DataFrame(response.json())
```
frgfm, line 302: not sure if that's on purpose, but those are not "all events", only the ones that haven't been acknowledged

app/main.py
```diff
@@ -408,6 +412,7 @@ def update_live_alerts_data(
             # - the storage component that contains the dictionary with detection frame URLs;
             # - the storage component that serves as source of truth for the list of already loaded alerts;
             # - the HTML div that stores the list of sites where an alert is live
+            print(live_alerts, dict_images_url_live_alerts, new_loaded_frames, sites_with_live_alerts)
```
frgfm, line 415: I think that's for debug, careful to remove those when cleaning the PR :+1:

frgfm (commented):
A few comments, but more importantly, tests are failing :/

app/main.py
Akilditu: True but that's only a naming convention

app/main.py
Akilditu: Yes, thanks for that

Akilditu: Done

frgfm (commented):
Thanks! I think there is only one typo left

frgfm (approved):
Thanks!

PR #81, opened 2022-07-04
refactor: Refactored project & updated dependencies

### LINKED ISSUE #35: [Content] Large content files should be removed from repo

### LINKED ISSUE #80: [404] Wrong behaviour for 404 errors
Currently if a user tries to access any other URL than the expected ones, it will build the monitoring screen. This should lead to a "404 Error" message.

frgfm:
This PR introduces the following modifications:
- updated dependencies after carefully testing them in Docker
- moved dependency specifiers to `pyproject.toml` (holding the version ranges) so that poetry can lock the production version (putting specific version into a `requirements.txt`)
- moved subURLS: `alert_screen` --> `minimal`, `dashboard_screen` --> `device_status`
- added a user warning for 404 errors
- added Sentry integration through Flask
- reorganized project hierarchy (cf. below)
- removed legacy section from README
- removes cameras.csv & historic_fires.csv
- updated headers & header verification job

The new project organization is as follows:
```
app
├── assets
│   ├── css
│   │   └── styles.css
│   ├── data
│   │   └── group_correspondences.json
│   └── images
│       ├── background.png
│       ├── favicon.ico
│       ├── pyro_alert_icon.png
│       ├── pyro_alert_off.png
│       ├── pyro_alert_on.png
│       ├── pyro_fire_logo.png
│       ├── pyro_oldfire_icon.png
│       ├── pyro_site_icon.png
│       └── up_arrow_png.png
├── components
│   └── navbar.py
├── config.py
├── main.py
├── pages
│   ├── device_status.py
│   ├── homepage.py
│   ├── risks.py
│   └── screen.py
├── services
│   ├── api.py
│   └── __init__.py
└── utils
    ├── alerts.py
    └── _utils.py
```

The folders & files are:
- assets: all the static files
- components: html/dsh components used to build pages
- config.py: same as before
- main.py: same as before
- pages: 1 file per app page
- services: same as before
- utils: other functions

One thing that will be left to do is reduce the size of main (moving callbacks or creating subfunctions of utils) because it's way too massive.

Closes #35 
Closes #80

Any feedback is welcome!


frgfm:
Deployed on Heroku, everything seems to be running smoothly :+1:

app/main.py
```diff
@@ -26,81 +26,65 @@
...
+        dsn=cfg.SENTRY_DSN,
+        release=cfg.VERSION,
+        server_name=cfg.SERVER_NAME,
+        environment="production" if isinstance(cfg.SERVER_NAME, str) else None,
+        integrations=[
+            FlaskIntegration(),
+        ],
+        traces_sample_rate=1.0,
+    )
+    logger.info(f"Sentry middleware enabled on server {cfg.SERVER_NAME}")
```
Akilditu, line 87: Could you explain how does sentry interacts with the platform ?

app/main.py
```diff
@@ -177,17 +161,20 @@
...
     thanks to the instantiation functions built in the various .py files.
     """
-    if pathname == "/alert_screen":
+    if len(pathname) == 0 or pathname == "/":
+        return Homepage()
+    elif pathname == "/screen":
         return AlertScreen()
-    elif pathname == "/dashboard_screen":
+    elif pathname == "/device_status":
         return DashboardScreen()
```
Akilditu, line 174: I understand path renaming here but shouldn't we match function name as well (e.g DashboardScreen) ?

app/components/navbar.py
```diff
@@ -70,7 +70,7 @@ def Navbar(dropdown=False):
                 ]
             )
         ],
-        href="alert_screen",
+        href="screen",
```
Akilditu, line 73: Not sure about this renaming ?

requirements.txt
```diff
@@ -1,18 +1,46 @@
...
+certifi==2022.6.15; python_version >= "3.7" and python_version < "4"
+charset-normalizer==2.1.0; python_version >= "3.7" and python_version < "4" and python_full_version >= "3.6.0"
+click==8.1.3; python_version >= "3.7" and python_full_version < "3.0.0" or python_full_version >= "3.3.0" and python_version >= "3.7"
+colorama==0.4.5; python_version >= "3.7" and python_full_version < "3.0.0" and platform_system == "Windows" or python_full_version >= "3.5.0" and python_version >= "3.7" and platform_system == "Windows"
+dash-bootstrap-components==0.13.1; (python_version >= "2.7" and python_full_version < "3.0.0") or (python_full_version >= "3.5.0" and python_version < "4")
+dash-core-components==1.17.1; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version < "4"
+dash-extensions==0.0.71
+dash-html-components==1.1.4; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version < "4"
+dash-leaflet==0.1.23
+dash-table==4.12.0; python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.5.0" and python_version < "4"
```
Akilditu, line 13: Thanks for update dash static versions. Will run some additional tests to make sure that everything is working as expected !

Akilditu (commented):
Hi @frgfm,

Thanks a lot for all those refactor / updates, left some small comment and will run some tests as well but everything's looks good overall :)

app/main.py
frgfm: It's experimental, but on the platform side, whenever there is a crash, it will send a sentry alert (similar to the API). However it's a bit obscure how to do it with dash, so this is experimental :+1:

app/main.py
frgfm: Agreed, I'm changing this as well :+1:

app/components/navbar.py
frgfm: perhaps "alert_minimal" ?

alert_screen sounds to me as if it could fit both the homepage & that screen

frgfm: or simply "minimal" ? (minimal / simple version of the full homepage)

requirements.txt
frgfm: This file is generated from the specifiers I put in `pyproject.toml`. So the process now is to :
- edit pyproject.toml
- it will generate poetry.lock & requirements.txt

frgfm:
ping @Akilditu

Issue #83, opened 2022-07-14
[credentials] Logged credentials aren't used for API interactions


frgfm:
There is a major problem regarding credentials:
- a user is indeed authenticated on the login screen using their personal creds
- but once logged, only the admin creds are used

The consequence is that the scopes are not respected: a user will see the devices of someone else. This needs to be fixed shortly!

pechouc:
Once we instantiate the API client with the credentials used for login, we can use it to load alerts for the first time but we cannot store the client Python object to update the alerts afterwards. The simplest solution could be to (i) use the login credentials to create the API client and load the alerts for the first time with it, (ii) store the associated token in a dedicated Dash component and (iii) use this token to authenticate the user in the next calls to the API. How does this sound to you?

frgfm:
@pechouc yes I agree!
The routes of the alert API were designed for this :) I have no clue how to handle the cache in dash though :sweat_smile:

pechouc:
Ok, I have started implementing this approach. Just a question however: how should we deal with the extinction of the token? 

One possibility would be to store the credentials used for logging (the same way we store the user's initial token) and use them to re-authenticate each time we have a 401 error. But it feels weird to put the username and password somewhere of course. 

Or maybe we can regularly refresh the token if the session keeps going? Adding a timer and updating the component that stores the token would be easy via Dash, but I am not sure of what would be the procedure to refresh the token.

frgfm:
> Just a question however: how should we deal with the extinction of the token?

Do you mean this? https://github.com/pyronear/pyro-platform/blob/master/app/main.py#L98-L100


> One possibility would be to store the credentials used for logging (the same way we store the user's initial token) and use them to re-authenticate each time we have a 401 error. But it feels weird to put the username and password somewhere of course.

I'm not sure I fully get what you mean, but we don't need to store the creds unless we need to refresh. So I suggest caching the token, and rerequest the user to enter the creds when the token expires? Otherwise we need to encrypt this somehow


> Or maybe we can regularly refresh the token if the session keeps going? Adding a timer and updating the component that stores the token would be easy via Dash, but I am not sure of what would be the procedure to refresh the token.

Either way, to refresh the token, you need the creds :sweat_smile:

pechouc:
Yes, this is indeed what I mean. 

And I don't think we want firemen to re-enter the credentials for the big minimal screen every hour, do we?

pechouc:
[...]

PR #85, opened 2022-07-16
Using login credentials

pechouc:
In relation to #83, this PR aims at using the user's credentials, used to log into the platform, for most of the following interactions with the database. This would allow to have multiple users on the platform, which is required with the Engie install. The tasks can be summarised as:

- [x] Storage of the user's token in a Dash component
- [x] Replacing client interactions with calls to the API using the user's headers
- [x] Dealing with the extinction of the token
- [ ] ~~Fetching the ID of the user's group after login (is there an appropriate route or client method?)~~
- [ ] ~~Adding coordinates and zoom for the Engine group in `app/assets/data/group_correspondences.json`~~


app/main.py
```diff
@@ -255,15 +255,19 @@ def update_live_alerts_data(
...
     client so as to identify live alerts, load the associated data and trigger their display on the platform. This doc-
     string should be completed but more details can be found in the comments below.
     """
 
+    if user_headers is None:
+        raise PreventUpdate
+
     # Fetching live alerts where is_acknowledged is False
-    response = api_client.get_ongoing_alerts()
+    response = requests.get(cfg.API_URL + "/alerts/ongoing", headers=user_headers)
```
frgfm, line 270: let's keep the API client, much safer (it will be updated if routes get updated)
We can update the token used by the client as such
```python
api_client.token = user_token
```

app/main.py
```diff
@@ -609,16 +621,24 @@ def manage_login_modal(n_clicks, username, password, login_storage, current_cent
...
+                {"center": [10, 10], "zoom": 3},
+                {"display": "none"},
+                dash.no_update,
+            ]
 
         else:
             # This is the route of the API that we are going to use for the credential check
             try:
-                Client(cfg.API_URL, username, password)
+                client = Client(cfg.API_URL, username, password)
```
frgfm, line 636: the variable isn't used afterwards, no need to keep it

frgfm (commented):
Small comment about token update :+1:

app/main.py
pechouc: It is used in the `return` statement just below (line 666).

app/main.py
pechouc: Good idea, thanks!

app/main.py
frgfm: oh my bad!

frgfm (commented):
Thanks!
We're on the good track, let's refresh the token now :+1:

Akilditu (commented):
Thanks a lot for this PR @pechouc here are some global thoughts : 

- Why storing user headers while `username` / `password` seem to be enough ?

> * [ ]  Dealing with the extinction of the token
I guess `api_client.refresh_token(`username`, `password`)` might do the job, no ? 
> * [ ]  Fetching the ID of the user's group after login (is there an appropriate route or client method?)
--> I can't see a link between `users` and `groups` but I can see links between `sites` and `groups` as well as `devices` and `owner_id` (`user_id` ?) 
@frgfm I might be missing something but this needs to be discussed
> * [ ]  Adding coordinates and zoom for the Engine group in `app/assets/data/group_correspondences.json`
--> We could stick to this for now and then try to dynamically deduce a midpoint regarding sites coordinates

frgfm:
> Why storing user headers while username / password seem to be enough ?

I agree for the creds but I think it was out of safety that this wasn't stored in the cache first. It's not exactly optimal to store unencrypted creds in the cache, but that might be our first step :man_shrugging: 

> --> I can't see a link between users and groups but I can see links between sites and groups as well as devices and owner_id (user_id ?)

Well in the API, group association was introduced for regional management. Being admin vs. user wasn't enough: as a user you can fetch your device, so if you try to get ALL devices, that will get you yours only. However if other user are in the same group as you are, you can read those devices s well :) 

So typically, what we want is:
- the admin to keep access to everything (Pyronear sys admin, not on the firefighters side)
- each region (Ardèche vs. Landes) to have its own group and have "users" scope

That will allow us to have (if needed) multiple users from the same regions sharing a similar data access authorization.

> Adding coordinates and zoom for the Engine group in app/assets/data/group_correspondences.json

Let's make this dynamic later, but updating this manually for now :+1:

pechouc:
Site markers shown on the map now depend on the credentials entered by the user. Typically, the Engie user can only see the corresponding sites and devices. 

The only issue that remains is the extinction of the token after 1 hour. For now, we refresh the token based on super admin credentials. As a consequence, after 1 hour, all sites and alerts are shown on the platform for any user. The simplest way to move forward while dealing with this issue would be to (i) store the credentials of the user and (ii) use them to refresh tokens. Of course, this is far from ideal but it can be implemented fairly easily.

Akilditu (commented):
Hi @pechouc, 

Thanks a lot for this final touch, everything seems to work just fine after some running some tests !

Only one matter regarding token expiration. Shouldn't we fallback to the user creds as you mentioned ? 
I understand that having to store username + pwd isn't ideal but I thing it's better than revealing all alerts + towers from super admin after 1h tokens expirations. 

I would recommend implementing the user creds patch waiting for a better solution handled by the API (e.g token refresh thanks to token instead of creds ?)

pechouc:
Great, thanks a lot!

> I would recommend implementing the user creds patch waiting for a better solution handled by the API

I agree with this temporary solution, I can quickly implement it during the weekend if we validate it.

> e.g token refresh thanks to token instead of creds ?

I am afraid that, if the token were enough to keep access to the API for an unlimited amount of time thanks to refreshing, this would ruin the initial purpose of the token expiration.

Maybe an alternative could be to condition the `refresh_token` action on a sort of private admin key, stored as a secret environment variable on Heroku, rather than on the credentials? This way, we could refresh the platform user's token, without taking the risk that someone outside the organisation could find an API token and use it indefinitely?

frgfm:
> But this discussion may of course fall outside the scope of this PR and we may want to open another issue if we agree that edits should rather be made on the API side.

From a security point of view, we cannot issue unexpiring token for users :/

frgfm (approved):
Thanks!

For this PR, would you mind attaching a screenshot illustrating the "user-only" objects that are accessed?

Also for testing purposes, in another PR, I'm starting to think we should move most functions to a local python package (so that we can test it outside of dash). That would make the app more robust and make the difference between backend & pure frontend features :+1:

frgfm:
> Sounds great! Do you think you could open a dedicated issue? Because I would have 1 or 2 questions to make sure I understand how this should be organised blush

@pechouc will do!

> This would keep token refresh safe as admin creds needed and would keep active user sessions for platform purposes

@Akilditu that means that anyone will be able to read the creds as they will be stored as instance attributes :/

pechouc:
Hello, I added the storage components for the user's credentials and used those for refreshing the token. I have tested it, leaving the platform open for an hour, and it seemed to work well!

Akilditu (approved):
Thanks @pechouc !

Shouldn't we still create an issue regarding this user creds storage ? 

As discussed we do not have a better option right now but this needs to be addressed right ?

pechouc:
I agree that this should be by no means definitive and thanks for opening the dedicated issue. Should we merge this for now though?

Akilditu:
> I agree that this should be by no means definitive and thanks for opening the dedicated issue. Should we merge this for now though?

Yes !

frgfm (approved):
Thanks! I imagine you have tested the PR to make sure it works? (after an hour)

pechouc:
@Akilditu has noticed an issue with the device status dashboard. The last commit ensures that:

- We do not manually filter the list of devices shown based on their ID (which was done to "hide" the test devices from users outside of Pyronear so far). This filtering operation is now executed implicitly, by using the user's headers to request the list of devices;
- We can deal with users that have no devices in their scope (building and displaying an empty table);
- The presence of some devices with missing values for the `last_ping` variable (which is the case of the two "Reolink" devices in the database) does not break the callback.

> Thanks! I imagine you have tested the PR to make sure it works? (after an hour)

Yes, this was tested.

app/main.py
```diff
@@ -1489,35 +1585,43 @@ def update_alert_frame_main(alert_frame_update_new_event, alert_frame_update_int
...
-    )
+    if not all_devices.empty:
+        sdis_devices = all_devices.sort_values(by="login")[["yaw", "lat", "lon", "login", "last_ping"]].copy()
 
-    sdis_devices["last_ping_hours_dif"] = sdis_devices["last_ping"].apply(
-        lambda x: (pd.to_datetime(x) - pd.to_datetime(datetime.utcnow().isoformat())).total_seconds() // 3600
-    )
+        # For some Reolink devices, it seems that the last ping is missing
+        # We add this line to prevent these cases from breaking the logic below
+        sdis_devices = sdis_devices.dropna(subset=["last_ping"]).copy()
```
Akilditu, line 1613: I checked and it seems that last_ping can't be `null` as there is a default value taking the creation timestamp

app/main.py
```diff
@@ -1489,35 +1585,43 @@ def update_alert_frame_main(alert_frame_update_new_event, alert_frame_update_int
...
-    response = requests.get("https://api.pyronear.org/devices/", headers=api_client.headers)
+    if user_headers is None:
+        raise PreventUpdate
+
+    response = requests.get("https://api.pyronear.org/devices/", headers=user_headers)
     # Check token expiration
     if response.status_code == 401:
-        api_client.refresh_token(cfg.API_LOGIN, cfg.API_PWD)
+        api_client.refresh_token(user_credentials["username"], user_credentials["password"])
         response = requests.get("https://api.pyronear.org/devices/", headers=api_client.headers)
```
Akilditu, line 1602: `/devices`/ route is actually on an admin-scope only. Now that we're using user creds I would recommend to use `/devices/my-devices` that returns user-scoped devices. 

To allow admin reporting (all devices) I would recommend : 

1. Use default `/devices/` route 
2. If not empty --> proceed for admin purposes
3. If empty then use `/devices/my-devices`

Akilditu (changes requested):
Thanks for dealing with the device data issue highlighted here #86 ! 

After running some tests it seems that moving on user creds API calls also impacts some routes such as `"admin-only"` `/devices/` one. 

I left some proposal as comments

app/main.py
pechouc: I do find a `None` last ping for two devices (IDs 31 and 32), which breaks the `datetime` conversion. The `software_hash` is also missing for these two. For now, the `drop_na` is still required for the callback to run 😢

app/main.py
pechouc: I completely missed this, thanks!

pechouc:
> I left some proposal as comments

Thanks a lot, I have either left a response here or implemented the changes. It now seems to work properly, but I have tested quickly and may have missed something.

app/main.py
Akilditu: Those are tests made by @fe51 that seem kinda buggy to me 😅 

Ok to let this safety then !

Akilditu (approved):
Thanks for the changes @pechouc, ok to merge ! :)

frgfm:
@Akilditu sure, but you have to login through Heroku !
Strangely, the traceback isn't available for higher level :/

It came from a MAC user on Safari

pechouc:
> It came from a MAC user on Safari

This definitely sounds like me 😅 I did run into some bugs while testing the PR locally, which is likely reflected in these errors. They should be solved now, but I won't have time today to fully confirm this...

PR #92, opened 2023-04-15
fix: Fixes interactions with alert API & style tool version constraints

frgfm:
This PR is meant to be merged before #91 to solve a few aspects:
- pinning the black version used by the CI
- doing the same for the mypy version
- updating the service name to a more appropriate one
- fixed the headers

Apart from codebase edits, I did:
- boot a dev bucket & API instance
- updated the secrets in this repo accordingly

Final note: it's "normal" that some CI jobs fail as the API url is not dynamic yet until #91 is merged


blenzi:
Thanks @frgfm ! Now the credentials are fixed, I re-ran the checks in #91 and they now pass
- The issue with mypy was fixed #91 
- I didn't ask before but why do you think it is better to stick to the previous version of black instead of moving to the one used in #91 (and sticking to that version) ?

frgfm:
My apologies for the late reply @blenzi :upside_down_face: 

Regarding black, we need to stick to one version. And the one I specified in this PR is the one that has been used on all our repos so I suggest staying on that one for harmonization sake :smiley:

blenzi (approved):
Fair enough! Thanks for the changes

PR #93, opened 2023-04-25
fix: Fixes some issue related to old YAW statements

Akilditu:
This PR aims at solving some issues since last API improvements such as:

- replacing old `yaw` fields by the new `azimuths` ones
- deleting old merges between alerts and devices to catch so called `azimuths` as they are available in both tables now
- force '/' before any hardcoded routes so that it soothly interact with `API_URL` ending with no '/'

Some thoughts before merging:

- Adjustments need to be done on API's config as conflicts happen between groups / owners / users / installations
- App not ready to run through Docker yet, catches some local API issue in that regards


blenzi (commented):
Thanks a lot @Akilditu ! Since you will need to run black to pass the remaining check, I took the chance to ask a small clarification and suggest a minor change that hopefully will improve clarity. Otherwise looks good to me!

frgfm (commented):
I agree with @blenzi review :)
Nice that we can avoid some pandas operations with the yaw --> azimuth update

PR #94, opened 2023-05-04
chore: use pyroclient from github

blenzi:
The latest release of pyroclient (0.1.2) is not compatible with the platform anymore (e.g. https://github.com/pyronear/pyro-api/pull/215). Use the version from github while waiting for a compatible release.

N.B.: using a version from github requires to have git to install it, so the image cannot be slim. Use `python:3.8` instead.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.8
```
frgfm, line 1: That usually saves us a decent amount of docker image size, is this mandatory?

frgfm (commented):
Thanks Bruno :)
Only need to merge main into this, and perhaps check the base docker image!

Dockerfile
blenzi: Unless we install git with apt. Not sure it will save us a lot of space. Once a new release of the API is available, we can move back to -slim

Akilditu (approved):
Everything looks fine to me. Thanks for adding this API client install through git as the package is not uptodate

MateoLostanlen:
Done in #101

PR #95, opened 2023-05-12
test: add requirements workflow

blenzi:
Add a new workflow `requirements` to check consistency of `poetry.lock` and `requirements.txt`


frgfm:
thanks Bruno :) Similarly to the API, I removed the duplicate specification system, so that we only use poetry for docker-based projects :+1: (cf. #97)

PR #111, opened 2023-10-17
display alerts from today

MateoLostanlen:
As requested by several users, I propose to display only the current day's alerts, as the previous days are useless for firefighters.

There are two options for accessing the history:
A dashboard page with history stats and access to images. 
The ability to specify a date range for alerts to be displayed.

In my opinion, the dashboard is the best option, let's just do it in another PR.

The PR is relatively simple: I filter the days that are not equal to today.

I fixed a small error with filtered_df at the same time

### DIFF
--- a/app/utils/alerts.py
+++ b/app/utils/alerts.py
@@ -288,9 +288,13 @@ def build_alerts_elements(images_url_live_alerts, live_alerts, map_style):
     alerts_markers_layer = dl.LayerGroup(children=alerts_markers, id="alerts_markers")
 
     # Building the alerts notification btn
-    nb_alerts = len(all_events)  # Number of unique events
+    today = pd.Timestamp.now().normalize()
+    today_alerts_df = all_events[all_events["created_at"] >= today]
+
+    # The rest of the code remains the same...
+    nb_today_alerts = len(today_alerts_df)  # Number of unique events from today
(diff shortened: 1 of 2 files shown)


app/utils/alerts.py
```diff
@@ -288,9 +288,13 @@ def build_alerts_elements(images_url_live_alerts, live_alerts, map_style):
     alerts_markers_layer = dl.LayerGroup(children=alerts_markers, id="alerts_markers")
 
     # Building the alerts notification btn
-    nb_alerts = len(all_events)  # Number of unique events
+    today = pd.Timestamp.now().normalize()
+    today_alerts_df = all_events[all_events["created_at"] >= today]
```
blenzi, line 292: Both look good but maybe using always the same type of comparison (`>= today` or `day == today`) would be clearer ?

blenzi (commented):
Thanks a lot for this addition @MateoLostanlen. Just a minor suggestion about the comparisons.

I agree that a separate functionality and ideally another dashboard that allows retrieving the images and information about older alerts (by date, site, etc) is required. At very least this PR must pass to allow it on the API side: https://github.com/pyronear/pyro-api/pull/277.

blenzi (approved):

PR #118, opened 2023-10-30
Draw bbox

MateoLostanlen:
The PR to display the prediction the firefighters wanted! 

I've added a button to toggle the prediction on or off for a better view in the case of a small fire.

I've removed the question "Is the alert relevant?", which is redundant with the question below, to free up space.

I'm going to update the devices with the latest version of engine so that they all send the prediction.

Screenshot from 2023-10-30 11-55-13

Screenshot from 2023-10-30 11-55-19

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -64,6 +64,7 @@ from utils.alerts import (
     build_sites_markers,
     display_alert_selection_area,
     past_ndays_live_events,
+    process_bbox,
     retrieve_site_from_device_id,
 )
 
@@ -117,6 +118,7 @@ app.layout = html.Div(
             id="store_live_alerts_data", storage_type="session", data=json.dumps({"status": "never_loaded_alerts_data"})
         ),
         dcc.Store(id="images_url_live_alerts", storage_type="session", data={}),
+        dcc.Store(id="images_boxes_live_alerts", storage_type="session", data={}),
         dcc.Store(id="last_displayed_event_id", storage_type="session"),
         dcc.Store(id="images_to_display_on_big_screen", data={"frame_URLs": "no_images"}, storage_type="session"),
         html.Div(id="alert_frame_update_new_event", style={"display": "none"}),
@@ -215,6 +217,7 @@ def change_layer_style(n_clicks=None):
     [
         Output("store_live_alerts_data", "data"),
         Output("images_url_live_alerts", "data"),
+        Output("images_boxes_live_alerts", "data"),
         Output("loaded_frames", "data"),
         Output("main_api_fetch_interval", "interval"),
         Output("sites_with_live_alerts", "children"),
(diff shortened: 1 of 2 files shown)


Akilditu (commented):
Hi @MateoLostanlen, thanks a lot for this PR, overall everything seems OK. 

Sometimes there is no bbox on detection frames (screenshot bellow), could you explain why ? 
Also in this case should we still display the frame as the ticked option seems to be confusing.

Capture d’écran 2023-10-31 à 09 37 11

MateoLostanlen:
hello @Akilditu ,

Pyro-engine combines 4 frames to send an alert in order to reduce FP rate. The box we display is the prediction of the model on each frame, if the model do not detect anything there is no box

app/main.py
```diff
@@ -1057,22 +1086,44 @@ def display_alert_modal(n_clicks):
...
-    When the user has opened an alert modal, he or she can choose the alert frame to view thanks to the slider. This
-    callback is triggered by a change in value of the slider and return the URL address of the frame to be displayed.
-    Like there is one alert modal per event, there is one alert slider per event, which allows to use MATCH here.
-    """
-    if slider_value is None:
+def select_alert_frame_to_display(slider_value, bbox_toggle_value, urls, bboxes_dict):
+    event_id = dash.callback_context.inputs_list[0]["id"]["index"]
+    # Check if the urls for the specific event_id exists
+    if slider_value is None or urls is None or event_id not in urls:
         raise PreventUpdate
```
Akilditu, line 1106: Since this last commit, after running some test it seems that that this addon raises PreventUpdate and then frame remains the same + there is no bbox drawn.

Akilditu (commented):
Also small thought when you delete code please try to let some comments about defs so that we can understand the purpose of it. 

I know they're quite long but but do not delete everything 😅

Akilditu (commented):
@MateoLostanlen everything seems fine now, one last thing : I understand the 4 frames needed for Pyro Engine but should we still display frames on which there is no box / detection ? 
I understand that this would somehow break the 'detection flow' that shows the fire evolution but I find it kinda weird to have random frames with and without boxes in between. Feels a bit like the model is messing things up ..

MateoLostanlen:
@Akilditu  I understand your concern about the consistency of the detection flow. However, it's important to consider that every frame, whether it has a detected box or not, contains valuable information about the presence or absence of fire at a given moment. Removing frames without detections could potentially discard subtle cues or indicators that are crucial for a comprehensive understanding of the fire's evolution. 

Firefighters rely on the full context of an evolving situation. Even frames where the AI does not detect fire can serve as negative data points that contribute to the overall pattern of the event. These 'empty' frames can act as confirmation that the fire has not spread to certain areas at specific times, which is just as important as knowing where the fire is active.

Therefore, keeping all frames ensures that firefighters get the complete picture, which is essential for their safety and the effectiveness of their response. It is better to err on the side of providing too much information rather than too little in such critical applications.

Akilditu (approved):
All good then, let's wait for user feedbacks about detection frames without bbox and this interpolation solution we may have !

PR #119, opened 2023-10-30
dl current image

MateoLostanlen:
As it is not possible to zoom in on the images, the firefighters have asked us to add the possibility of downloading the image for better viewing

image


Akilditu (approved):
Thanks for this enhancement !

PR #122, opened 2023-10-31
force alert list update

MateoLostanlen:
Before the PR #113, the alert list was not updated correctly, so this problem was not visible. When an alert is acknowledged, it's essential to update the alert list, otherwise you'll be trying to access components that no longer exist


Akilditu (approved):
Hi Mateo, thanks for this fix !

PR #123, opened 2023-11-03
ci: Removes dangling images on remote server

frgfm:
This PR ensures docker images are cleaned upon deployment on remote server.

Any feedback is welcome!


MateoLostanlen (commented):
Perfect exactly what we needed thank you

Small question though, you're sure about "docker rmi -f" we don't want to force right?

Akilditu (approved):
Hi @frgfm, thanks for this 👍

PR #126, opened 2023-11-07
ci: fix typo

Akilditu:
This PR aims at correcting misspelling in docker compose command within deploy_dev CI job. 

Thanks for reviewing !

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -37,6 +37,6 @@ jobs:
           key: ${{ secrets.SSH_DEPLOY_DEV }}
           script: |
             docker pull pyronear/pyro-platform:latest
-            cd pyro-platform-prod && docker compose down && docker-compose -f docker-compose-prod.yml up -d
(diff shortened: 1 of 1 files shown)


MateoLostanlen (approved):
Thanks !

This is not a misspelling both docker-compose and docker compose exist:

The docker compose (with a space) is a newer project to migrate compose to Go with the rest of the docker project. This is the v2 branch of the [docker/compose](https://github.com/docker/compose/tree/v2) repo. It's been first introduced to Docker Desktop users, so docker users on Linux didn't see the command. In addition to migrating to Go, it uses the compose-spec, and part of the rewrite may result in behavior differences.

The original python project, called docker-compose, aka v1 of docker/compose repo, has now been deprecated and development has moved over to v2. To install the v2 docker compose as a CLI plugin on Linux, supported distribution can now install the docker-compose-plugin package. E.g. on debian, I run apt-get install docker-compose-plugin.

PR #129, opened 2023-11-07
use image prune

### LINKED ISSUE #128: ci: Issue with rmi unused and dangling images
Hi @frgfm there is an error caused deploy-dev job on release as you can see [here](https://github.com/pyronear/pyro-platform/actions/runs/6783271093/job/18437235963) 

Trying to delete the current image running and then causing -->` err: Error response from daemon: conflict: unable to delete a8d09517dc80 (cannot be forced) - image is being used by running container 7bd1a0897dd8`

Shouldn't we use something like `docker image prune -a` ?

MateoLostanlen:
fix #128 as recommended by @Akilditu

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -39,4 +39,4 @@ jobs:
             docker pull pyronear/pyro-platform:latest
             cd pyro-platform-prod && docker compose down && docker compose up -d
             docker inspect -f '{{ .Created }}' $(docker compose images -q frontend)
-            docker rmi -f $(docker images -f "dangling=true" -q)
+            docker image prune -a --force


.github/workflows/push.yml
```diff
@@ -39,4 +39,4 @@ jobs:
             docker pull pyronear/pyro-platform:latest
             cd pyro-platform-prod && docker compose down && docker compose up -d
             docker inspect -f '{{ .Created }}' $(docker compose images -q frontend)
```
Akilditu, line 41: I'm not sure of the relevance of this then :/

We could delete it as well I guess

Akilditu (approved):
Let's go !

frgfm (commented):
Mmmmh are you positive this is a good idea?

The force prune will delete all the images I believe, and since it happens after the pull, it also deletes the image (or try to) being used. (same goes for the other images being used like postgres & traefik)

Also, if you have two versions (let's say we do a major release). It's better to rollback and so if you have tagged images, it's all good.

The previous design was specifically making sure it deletes the previous "latest" docker image. But if you have other tagged, it kept them.

Akilditu:
> Mmmmh are you positive this is a good idea?
> 
> The force prune will delete all the images I believe, and since it happens after the pull, it also deletes the image (or try to) being used. (same goes for the other images being used like postgres & traefik)
> 
> Also, if you have two versions (let's say we do a major release). It's better to rollback and so if you have tagged images, it's all good.
> 
> The previous design was specifically making sure it deletes the previous "latest" docker image. But if you have other tagged, it kept them.

This actually seems to work fine (as you can check [here](https://github.com/pyronear/pyro-platform/actions/runs/6786077095/job/18445861053)

Also as mentioned in #128 the last command was actually trying to del current image

MateoLostanlen:
@frgfm the force here is needed to make it work in a ci : https://docs.docker.com/engine/reference/commandline/image_prune/ it's not to enforce it on all images

PR #130, opened 2023-11-07
ci: adding right config file

Akilditu:
This PR aims at adding the right config compose file for docker compose down and up !

Thanks for the review :)


MateoLostanlen (approved):
thanks !

PR #131, opened 2023-11-15
ci: fix dev auto push

Akilditu:
This PR aims at updating deploy_dev CI job with cleaned VM up to date satus and folders

Thanks for reviewing !


MateoLostanlen:
hi ! @Akilditu, maybe we can delete [docker-compose-prod.yml](https://github.com/pyronear/pyro-platform/blob/main/docker-compose-prod.yml) and [docker-compose-prod.yml](https://github.com/pyronear/pyro-platform/blob/main/docker-compose-dev.yml) if we are not using them then. Or it is used on prod VM ?

Akilditu:
> hi ! @Akilditu, maybe we can delete [docker-compose-prod.yml](https://github.com/pyronear/pyro-platform/blob/main/docker-compose-prod.yml) and [docker-compose-prod.yml](https://github.com/pyronear/pyro-platform/blob/main/docker-compose-dev.yml) if we are not using them then. Or it is used on prod VM ?

dev.yml is needed for dev using `make run_dev`. Deleted the prod though :)

MateoLostanlen (changes requested):
Ok looks good then, just delete your notebook please

MateoLostanlen (approved):

PR #152, opened 2024-06-05
Improve development environment

RonanMorgan:
Small fixes in order to make this project work with the pyro-devops development environment 
=> var env update
=> add instruction in the readme
=> add logging in the stdout


.env.example
```diff
@@ -0,0 +1,7 @@
+API_URL=http://localhost
+API_LOGIN=dummy_login
+API_PWD='dummy&P@ssw0rd!'
+
+SENTRY_DSN=
+DEBUG=
+LOGIN=
```
MateoLostanlen, line 7: missing blank line

Dockerfile
```diff
@@ -20,3 +20,5 @@ RUN apt update \
 
 # copy project
 COPY app/ /usr/src/app/
+
+CMD python index.py --host 0.0.0.0 --port 8050
```
MateoLostanlen, line 24: Not sur we need this it's already In the docker compose

MateoLostanlen: also missing blank line

MateoLostanlen (commented):
Hi Ronan, congratulations on your first PR on the platform. Just a few comments

Dockerfile
RonanMorgan: I like to have CMD or entrypoint directly in the Dockerfile to be able to use the container without the docker-compose , is it ok ?

=> I thought ruff would fix the blank line, I am gonna do it

MateoLostanlen: yes not problem for new just wanted to be sure that it was made on purpose

.env.example
RonanMorgan, line 7:
```suggestion
LOGIN=

```

RonanMorgan:
```suggestion
LOGIN=
 
```

Dockerfile
RonanMorgan, line 24:
```suggestion
CMD python index.py --host 0.0.0.0 --port 8050

```

PR #165, opened 2024-07-18
Use new event endpoint with media_url

RonanMorgan:
Instead of doing a loop on event id in order to get s3 url, the API is sending all the url with the event objects


MateoLostanlen:
done in #166

PR #166, opened 2024-09-05
New data model

MateoLostanlen:
There are a lot of changes in this PR, but it's quite simple: we simply adapt to the new format returned by the api for get_unacknowledged_events.

This new format combines alerts and events in a table with the urls already available, simplifying the code and speeding up loading time considerably.


app/callbacks/display_callbacks.py
```diff
@@ -268,11 +225,8 @@ def update_image_and_bbox(slider_value, alert_data, media_url, alert_list):
     images = []
     boxes = []
     for _, alert in alert_data.iterrows():
-        event_id = str(alert["event_id"])
-        media_id = str(alert["media_id"])
-        if event_id in media_url and media_url[event_id].get(media_id, "").strip():
-            images.append(media_url[event_id][media_id])
-            boxes.append(alert["processed_loc"])
+        images.append(alert["media_url"])
+        boxes.append(alert["processed_loc"])
```
frgfm, line 231: I know it doesn't sound obvious, but this will be faster & require less RAM:
```python
images, boxes = zip(*[(alert["media_url"], alert["processed_loc']) for _, alert in alert_data.iterrows()]) 
```

app/utils/display.py
```diff
@@ -184,7 +184,7 @@ def create_event_list_from_df(api_events):
             id={"type": "event-button", "index": event["id"]},
             children=[
                 html.Div(
-                    f"{event['device_name']}",
+                    f"{event['device_login'][:-2].replace('_', ' ') + ' - ' + str(int(event['device_azimuth'])) + '°'}",
```
frgfm, line 187: better to use f-string feature since you started with it:
```python
f"{event['device_login'][:-2].replace('_', ' ')} - {int(event['device_azimuth'])}°"
```

frgfm (dismissed):
I'm probably not the best for a plotly dash review, since I haven't checked the latest evolutions of the platform but LGTM. I added a few improvement comments for speed/memory/maintenance

app/callbacks/display_callbacks.py
MateoLostanlen: thnaks !

PR #167, opened 2024-09-11
feat: updated dockerfile to perform poetry installation + requirements automatically

Zougouda:
The previous version implied to already have a python + poetry installed on the host machine. Which was not very "dockery". I moved the required steps to install all of it in the dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,25 @@
 FROM python:3.9-slim
 
-# set work directory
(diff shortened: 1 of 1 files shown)


MateoLostanlen (changes requested):
Hello @Zougouda , thanks a lot for this pr. Sorry for the late review. I like the change, please can you just adapt the makefile ?

PR #169, opened 2024-11-06
Zoom on alert image

vanderlindenma:
Add the ability to zoom into main pictures.

https://github.com/user-attachments/assets/45610090-59e5-4ef6-983b-83f6ffa2bfe3

### DIFF
--- a/app/callbacks/display_callbacks.py
+++ b/app/callbacks/display_callbacks.py
@@ -184,8 +184,9 @@ def update_display_data(event_id_on_display, local_alerts):
 
 @app.callback(
     [
-        Output("image-container", "children"),  # Output for the image
-        Output("bbox-container", "children"),  # Output for the bounding box
+        Output("main-image", "src"),  # Output for the image
+        # Output("bbox-container", "children"),  # Output for the bounding box
+        Output("bbox-positioning", "style"),
         Output("image-slider", "max"),
     ],
     [Input("image-slider", "value"), Input("alert_on_display", "data")],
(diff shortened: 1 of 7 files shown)


app/callbacks/display_callbacks.py
```diff
@@ -209,19 +207,14 @@ def update_image_and_bbox(slider_value, alert_data, alert_list):
     - int: Maximum value for the image slider.
     """
     img_src = ""
-    bbox_style = {}
-    bbox_divs: List[html.Div] = []  # This will contain the bounding box as an html.Div
+    bbox_style = {"display": "none"}  # Default style for the bounding box
```
vanderlindenma, line 210: Same spirit as https://github.com/pyronear/pyro-platform/pull/169/files#r1831286489, if you need to temporarily hide an element, you can use this style attribute to do so instead of removing it from the DOM altogether and re-adding it when it needs to be shown again.

app/pages/homepage.py
```diff
@@ -23,20 +23,48 @@ def homepage_layout(user_headers, user_credentials):
...
+                                                        style={"display": "none"},
+                                                        children=[
+                                                            html.Div(
+                                                                id="bbox-styling",
+                                                                style={
+                                                                    "border": "2px solid red",
+                                                                    "height": "100%",
+                                                                    "width": "100%",
+                                                                    "zIndex": "10",
+                                                                },
```
vanderlindenma, line 60: This breakdown between `bbox-positioning` and `bbox-styling` elements is a hacky way to divide between those styling aspects that are changed by Dash updates, and those that are changed by JavaScript.

In this case, positioning is changed by Dash but JS needs a way to access `border` separately to reduce the width of the bbox as the picture is zoomed onto (in a way that sticks regardless of changes to the positioning).

app/utils/sites.py
```diff
@@ -12,7 +12,7 @@
 
 
 def get_token(api_url: str, login: str, pwd: str) -> str:
-    response = requests.post(f"{api_url}/login/access-token", data={"username": login, "password": pwd}, timeout=3)
+    response = requests.post(f"{api_url}/login/access-token", data={"username": login, "password": pwd}, timeout=10)
```
vanderlindenma, line 15: The API login was timing out on me every other time with only a 3 second wait (perhaps because I am in the U.S.?).

MateoLostanlen: it might be the case I have difficulties in Chile too, we can update this but please do it in an other PR it's not the topic here :)

app/callbacks/display_callbacks.py
```diff
@@ -184,8 +182,8 @@ def update_display_data(event_id_on_display, local_alerts):
 
 @app.callback(
     [
-        Output("image-container", "children"),  # Output for the image
-        Output("bbox-container", "children"),  # Output for the bounding box
+        Output("main-image", "src"),  # Output for the image
+        Output("bbox-positioning", "style"),
```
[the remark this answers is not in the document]
MateoLostanlen: good catch ! thanks for this update I make more sens

app/callbacks/display_callbacks.py
MateoLostanlen: same answer then :) thanks !

MateoLostanlen (approved):
LGTM !

PR #170, opened 2024-11-07
Increase API login timeout

vanderlindenma:
### WHAT

The API login was timing out on me every other time with only a 3 second wait (perhaps because I am in the U.S.?).


MateoLostanlen (approved):
Same issue in Chile, thanks for the PR !

PR #172, opened 2024-11-08
Add loading spinner on login

vanderlindenma:
https://github.com/user-attachments/assets/bbc7a764-1cdc-4a93-b110-472d3a5dcd70

### DIFF
--- a/app/callbacks/data_callbacks.py
+++ b/app/callbacks/data_callbacks.py
@@ -56,9 +56,11 @@ def login_callback(n_clicks, username, password, user_headers):
     This function is triggered when the login button is clicked. It verifies the provided username and password,
     attempts to authenticate the user via the API, and updates the user credentials and headers.
     If authentication fails or credentials are missing, it provides appropriate feedback.
+    After login succeeds and while data required to boot the dashboard is being fetched from the API,
+    the login form is hidden and a spinner is displayed.
 
     Returns:
-        dash.dependencies.Output: Updated user credentials and headers, and form feedback.
+        dash.dependencies.Output: Updated user credentials and headers, and form feedback + styles to hide/show login and loading spinners.
     """
     input_style_unchanged = {"width": "250px"}
     connection_button_style_unchanged = {}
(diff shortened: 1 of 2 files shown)


app/callbacks/data_callbacks.py
```diff
@@ -52,12 +57,28 @@ def login_callback(n_clicks, username, password, user_headers):
...
     If authentication fails or credentials are missing, it provides appropriate feedback.
+    After login succeeds and while data required to boot the dashboard is being fetched from the API,
+    the login form is hidden and a spinner is displayed.
 
     Returns:
-        dash.dependencies.Output: Updated user credentials and headers, and form feedback.
+        dash.dependencies.Output: Updated user credentials and headers, and form feedback + styles to hide/show login elements and loading spinners.
     """
+    input_style_unchanged = {"width": "250px"}
+    empty_style_unchanged = {"": ""}
```
vanderlindenma, line 67: Would be cleaner to user `{}` instead, but then `mypy` complains because it wants to know the type of `empty_style_unchanged` and it seems python does not infer types for empty dictionaries => using `{"": ""}` seems simpler than bringing in a whole typing library to handle that error only (`from typing import Dict, Optional ... empty_style: Dict[str, Optional[str]] = {}`).

app/callbacks/data_callbacks.py
```diff
@@ -80,12 +110,26 @@ def login_callback(n_clicks, username, password, user_headers):
...
+                return (
+                    dash.no_update,
+                    dash.no_update,
+                    form_feedback,
+                    input_style_unchanged,
+                    input_style_unchanged,
+                    empty_style_unchanged,
+                    empty_style_unchanged,
+                    hide_element_style,
+                )
```
vanderlindenma, line 132: @MateoLostanlen @fe51 @RonanMorgan Another example where `Dash` does not scale well. In a "sane" 😉 web framework:

1. You shouldn't have to repeatedly include placeholders for things you don't want to update: Just pass a map/dict of what you _do_ want to change and the framework should know the rest should be left untouched.
2. You shouldn't have to go back and forth between the parameters in the head of your function and the return to match the order of the declared outputs and with the order of the returns => just return a map each time.

MateoLostanlen (approved):
Thanks for this !

PR #173, opened 2024-11-08
Translate text

vanderlindenma:
### WHAT

Translate text in dashboard through button click, both on login and on dashboard page.

Screenshot 2024-11-12 at 9 33 08 AM

Screenshot 2024-11-12 at 9 33 32 AM

### NOTE

For changes to the translations, please submit follow up PR rather than requesting changes here (use this PR to assess viability of technical translation solution).


MateoLostanlen (approved):
Thanks !

PR #175, opened 2024-11-12
Add SAFE_DEV_MODE env that disables post requests to API

vanderlindenma:
### WHAT

Add SAFE_DEV_MODE env that disables post requests to API


MateoLostanlen (approved):
thanks !

PR #177, opened 2024-12-17
Reword: "location" to "camera location"

fe51:
In the right-hand side panel, the name of the location field is ambiguous. 
You might think that it refers to the location of the smokes/wildfire, but in fact it refers to the position of the camera.

Hence, this PR introduces the following modifications : 
- Updates field value (in french & spanish) : from "localisation" to "position caméra"
- Updates variable names & def in code to avoid confusion

Happy to discuss it ! :)

image


MateoLostanlen (approved):
Thanks

PR #191, opened 2025-02-13
prevent empty

MateoLostanlen:
We have the following bug:

```
KeyError: 'bboxes'
(6 additional frame(s) were not displayed)
...
  File "/usr/src/app/callbacks/data_callbacks.py", line 230, in load_detections
    detections["processed_bboxes"] = detections["bboxes"].apply(process_bbox)
```

 I can't reproduce it, I imagine it's because of an empty detection but I'm testing the existence of the column just in case


fe51 (approved):
Wierd, we might need to investigate further, but that might fix this error, LGTM !

PR #195, opened 2025-02-25
Bug : Fix local timezone display

fe51:
With the migration to the new data model, a regression has been introduced in the display of times. 

Times are now displayed in UTC, whereas the expected behaviour is to display them according to the camera's time zone. 
(We could discuss having a display relative to the location of the cameras rather than the user, but this is not currently a problem).

This PR fixes this bug. 

It introduces the following modifications : 
- Removed unused and outdated datetime conversion function
- Added conversion function to display utils
- Added datetime conversion when displayed

Happy to discuss it

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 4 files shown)


MateoLostanlen (approved):
Thanks, sorry for this error introduced during data model change

PR #261, opened 2025-08-13
Renamed tab name

fe51:
Currently, when users bookmark the platform in their browser or create a shortcut on their desktop, the default name is ‘Clickable Stream’ (see screenshot below)

Hence, this small PR modifies the wording to 'Pyronear' which is more explicit and avoid further renaming or ambiguity !

happy to discuss it !

Capture d’écran 2025-08-13 à 17 45 43


MateoLostanlen (approved):
good catch thanks

